PR #290, opened 2025-03-05
287 bugui ping service window selection broken

### LINKED ISSUE #287: bug(ui): ping service window selection broken
selecting 1h/6h/24h does nothing and needs to be removed, we only keep 100~ metrics for this and its all in memory.

mfreeman451:


mfreeman451 (commented):
lgtm

PR #342, opened 2025-03-13
combining snmp charts

### LINKED ISSUE #293: feat(snmp): combine multiple collections
if we have multiple counters being collected we should be able to configure a graph to show a combined view, layered chart stuff, of say both an in/out bytes counter.

mfreeman451:


web/src/components/SNMPDashboard.jsx
```diff
@@ -207,6 +323,34 @@ const SNMPDashboard = ({ nodeId, serviceName, initialData = [], initialTimeRange
...
+        }
+        if (metric.includes('Out')) {
+            return {
+                stroke: '#22c55e',
+                fill: '#86efac'
+            };
+        }
+
+        // Fallback colors
+        const colorPalette = [
```
mfreeman451, line 342: should be configurable

mfreeman451 (commented):
lgtm

PR #442, opened 2025-03-28
Feat/agent sweeper kv integration 02

mfreeman451:


cmd/agent/main.go
```diff
@@ -40,36 +40,32 @@ func run() error {
...
+		return fmt.Errorf("failed to load config: %w", err)
 	}
 
-	// Create agent server
-	server, err := agent.NewServer(ctx, cfg.CheckersDir, &agent.ServerConfig{
-		ListenAddr: cfg.ListenAddr,
-		Security:   cfg.Security,
-	})
+	// Create agent server with checkers_dir from config
+	server, err := agent.NewServer(ctx, "/etc/serviceradar/checkers", &cfg)
```
mfreeman451: should fix this

cmd/agent/main.go
```diff
@@ -40,36 +40,32 @@ func run() error {
...
+	server, err := agent.NewServer(ctx, cfg.CheckersDir, &cfg)
 	if err != nil {
 		return fmt.Errorf("failed to create server: %w", err)
 	}
 
 	// Create server options
 	opts := &lifecycle.ServerOptions{
 		ListenAddr:        server.ListenAddr(),
-		ServiceName:       cfg.ServiceName,
+		ServiceName:       "AgentService",
```
mfreeman451, line 64: not sure if this is a good idea, but going to leave it in here for now.

mfreeman451 (commented):
lgtm

cmd/agent/main.go
mfreeman451: updated

PR #592, opened 2025-04-12
feat(checker): zfs tools

### LINKED ISSUE #589: host-based metrics - gRPC checker
https://blog.guillaume-gomez.fr/articles/2021-09-06+sysinfo%3A+how+to+extract+systems%27+information

[sysinfo](https://docs.rs/sysinfo/latest/sysinfo/)

If we have a system with a high usage, need to try and identify what process or maybe what file/directory is causing the problem and surface that. Simply identifying systems with high usage is not enough, we need to tell the user WHY/what/how, if we can.

mfreeman451:
https://lib.rs/crates/shock
https://github.com/jimsalterjrs/sanoid


mfreeman451 (commented):
lgtm

PR #679, opened 2025-04-25
676 bugpackaging need to set api key

### LINKED ISSUE #676: bug(packaging): need to set API key
api key says CHANGEME after new install

mfreeman451:


mfreeman451 (commented):
lgtm

PR #707, opened 2025-04-29
Feat/proton adapter

### LINKED ISSUE #706: feat(poller): rewrite in rust
We are going to be updating the poller so that we can direct data to timeplus proton instead of directly to the `core` (API) service and thought this was a perfect time to just re-write the poller service in rust, given how simple it the golang implementation was.

mfreeman451:


cmd/poller-ng/src/adapter.rs
```diff
@@ -0,0 +1,596 @@
...
+                        service.service_name,
+                        processor.name(),
+                        e
+                    );
+                    }
+                }
+            }
+        }
+
+        // Always forward to core if configured, even after Proton processing
```
mfreeman451, line 498: need to revisit this later.. we probably dont always want to forward to the core when proton is enabled.

cmd/poller-ng/src/adapter.rs
```diff
@@ -0,0 +1,596 @@
...
+            error!("Error processing services for poller {}: {}", poller_id, e);
+        }
+
+        // Always return success to prevent unnecessary retries
+        Ok(Response::new(proto::PollerStatusResponse { received: true }))
+    }
+}
+
+fn get_batch_size(config: &Config) -> usize {
+    config.batch_size.unwrap_or(20) // Default to 20 concurrent checks per agent
```
mfreeman451, line 595: this should be configurable

mfreeman451 (commented):
lgtm

PR #708, opened 2025-05-02
Bug/core metrics db

mfreeman451:


pkg/core/server.go
```diff
@@ -810,15 +1021,15 @@ func (s *Server) processSNMPMetrics(pollerID string, details json.RawMessage, ti
...
-
-				continue
-			}
+			timeseriesMetrics = append(timeseriesMetrics, metric)
 		}
 	}
 
+	// Buffer SNMP timeseriesMetrics
+	s.bufferMu.Lock()
+	s.metricBuffers[pollerID] = append(s.metricBuffers[pollerID], timeseriesMetrics...)
```
mfreeman451, line 1030: not clear we need to be doing this, might come back and revisit it

pkg/db/db.go
```diff
@@ -7,419 +7,470 @@
...
 
-	if rowsAffected == 0 {
-		return sql.ErrNoRows
+	// Update pollers table
+	if err := db.insertPollerStatus(ctx, status); err != nil {
+		return fmt.Errorf("failed to update poller status: %w", err)
+	}
+
+	// Log to poller_history
+	if err := db.insertPollerHistory(ctx, status); err != nil {
```
mfreeman451: this looks dumb.. why are we inserting the same data into two different places?

mfreeman451 (commented):
lgtm

pkg/core/server.go
mfreeman451: addressed

pkg/db/db.go
mfreeman451: addressed, we are going to keep writing to the `pollers` table since it doesnt actually store a lot of data, and the performance hit from the pollers table is around .2 ms compared to 1-2ms to get it out of the `pollers_history` table, so 10-20x difference. we modified the updatepollers function so that it only updates the table when the status has changed as well

PR #861, opened 2025-05-29
Update/mapper polling interval fix

mfreeman451:


pkg/mapper/snmp_polling.go
```diff
@@ -91,15 +91,13 @@ const (
 	// oidCdpCacheDevicePort = ".1.3.6.1.4.1.9.9.23.1.2.1.1.7"
 	// oidCdpCacheAddress    = ".1.3.6.1.4.1.9.9.23.1.2.1.1.4"
 
-	// Default concurrency settings
-	defaultConcurrency = 10
-	defaultMaxIPRange  = 256 // Maximum IPs to process from a CIDR range
+	defaultMaxIPRange = 256 // Maximum IPs to process from a CIDR range
```
mfreeman451, line 94: that should be configurable

mfreeman451 (commented):
lgtm

pkg/mapper/snmp_polling.go
mfreeman451: will address later, we can leave this in now until someone complains, otherwise its a decent safeguard

PR #902, opened 2025-06-12
900 bugdb need to migrate proton tables for time series metrics to support new fields

### LINKED ISSUE #900: bug(db): need to migrate proton tables for time series metrics to support new fields

mfreeman451:


mfreeman451 (commented):
lgtm

PR #911, opened 2025-06-13
898 featarmis north bound integration with sweep results to armis

### LINKED ISSUE #898: feat(armis): north-bound integration with sweep results to armis
- [x] #913

### LINKED ISSUE #913: Update auth middleware
Update auth middleware so it accepts X-API-Key or bearer token, right now it requires both and that is a bit awkward for api-driven workflows.

mfreeman451:


mfreeman451 (commented):
lgtm

PR #974, opened 2025-06-19
Handle batch append errors correctly

mfreeman451:
## Summary
- fix device batch insertion logic to abort and return errors when append fails
- explicitly define columns for device insert to avoid argument mismatch

## Testing
- `go test ./...`

------
https://chatgpt.com/codex/tasks/task_e_68535cf6f9d08320ab2837b6cbb1ab39


mfreeman451 (commented):
lgtm

PR #982, opened 2025-06-19
Fix netbox source defaults

mfreeman451:
## Summary
- respect source-level agent_id & poller_id in sync
- include defaults when creating integrations
- emit devices from netbox with correct IDs and metadata
- test default resolution for integrations

## Testing
- `go test ./...`

------
https://chatgpt.com/codex/tasks/task_e_6853ae55d8708320aad7186160e717a4


mfreeman451 (commented):
lgtm

PR #983, opened 2025-06-19
Fix Netbox sweep config key generation

mfreeman451:
## Summary
- write sweep config using the configured agent ID instead of server name
- warn and skip when agent ID is missing

## Testing
- `go test ./...`

------
https://chatgpt.com/codex/tasks/task_e_6854422ece408320807cf60efd8bc1a0


mfreeman451 (commented):
lgtm

PR #997, opened 2025-06-20
Fix device ID generation to use partition

mfreeman451:
## Summary
- add partition column to sweep_results
- build `device_id` using `partition` and `ip`
- track partition in SweepResult model and integrations
- plumb partition through core server discovery logic

## Testing
- `go vet ./...` *(fails: unreachable code in generated parser)*
- `go test ./...`

------
https://chatgpt.com/codex/tasks/task_e_6854ee2dfa948320b2775955cc49e8e4


mfreeman451 (commented):
lgtm

PR #1013, opened 2025-06-21
Return alternate IPs as metadata array

mfreeman451:
## Summary
- roll back protobuf change and drop wrappers proto
- keep alternate IP addresses under a single `alternate_ips` metadata key
- store the metadata value as a JSON array when merging devices
- update polling logic and tests to use new helper

## Testing
- `go vet ./...` *(fails: unreachable code in generated parser files)*
- `go test ./...`

------
https://chatgpt.com/codex/tasks/task_e_6855f07579888320837069d2d5a82e4f

### DIFF
--- a/pkg/mapper/utils.go
+++ b/pkg/mapper/utils.go
@@ -18,6 +18,7 @@ package mapper
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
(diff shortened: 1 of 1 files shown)


mfreeman451 (commented):
lgtm

PR #1014, opened 2025-06-21
994 featcore support for partitions

### LINKED ISSUE #994: feat(core): Support for partitions
We are currently building device ids like this:

* ip-address:agent_id:poller_id

Besides the obvious issue that would present itself with IPv6 addressing, if we find the same device but it is being accessed through a different agent_id or poller_id, it will show up as a duplicate.

We need to introduce the concept of partitions to replace the current device_id naming strategy, we could probably just do "partition:ip address" and be able to achieve the same goal, and also support IPv6 by looking at everything after the first `:`.

Maybe the most obvious place to configure the 'partition' ID, would be at the poller level, where we define the `partition` in the poller.json config, and then we enrich messages we receive from the agent before we send them up to the serviceradar-core for further processing.

mfreeman451:


mfreeman451 (commented):
lgtm

PR #1026, opened 2025-06-21
875 featmapper UI add support for paginated query api 2

### LINKED ISSUE #875: feat(mapper-ui): add support for paginated query API
we've added support in the query API for pagination, need to add in support in the UI now

mfreeman451:


mfreeman451 (commented):
lgtm

PR #1033, opened 2025-06-21
Add multi-rule support to zen consumer

mfreeman451:
## Summary
- allow specifying `decision_keys` so multiple rules can be evaluated
- output results as CloudEvents
- watch new rules correctly
- let `zen-put-rule` specify a rule name when inserting
- document new options and provide example CEF rule

## Testing
- `cargo test --manifest-path cmd/consumers/zen/Cargo.toml`

------
https://chatgpt.com/codex/tasks/task_e_685642a4f1688320a26f8010aaefd7d9


mfreeman451 (commented):
lgtm

PR #1039, opened 2025-06-22
924 feat add support for gorules zen engine

### LINKED ISSUE #924: feat: add support for gorules zen engine
We now are ingesting SNMP traps and Syslog messages. They currently get sent to the NATS JetStream server (serviceradar-nats) where they are not being processed. 

We need the ability to act on these events, for instance, if we see a critical trap or a syslog event, we should be able to do something with them. 

https://github.com/gorules/zen 

Integrating the `gorules zen` engine into serviceradar gives us the ability to handle these events, using rust. 

The server needs to be able to be distributed and scale horizontally.

mfreeman451:


mfreeman451 (commented):
lgtm

PR #1054, opened 2025-06-22
1045 feat need database writer consumer 2

### LINKED ISSUE #1045: feat: need database writer consumer
Needs to listen to NATS JetStream "events.syslog.processed" and "events.snmp.processed" and then write messages to clickhouse in a new events table aligned with the CloudEvents event message format, and should use `async-nats` rust crate.

mfreeman451:


mfreeman451 (commented):
lgtm

PR #1058, opened 2025-06-22
Fix zen consumer config path

mfreeman451:
## Summary
- update service file to load config from /etc/serviceradar/consumers
- adjust Dockerfile to copy config into consumers directory
- create consumers dir in postinstall and package script
- update rpm spec and components.json with new config path
- fix leftover shell prompts in scripts

## Testing
- `make test` *(fails: signal interrupt)*

------
https://chatgpt.com/codex/tasks/task_e_685853aae6e0832093b2ea2721a391b3


mfreeman451 (commented):
lgtm

PR #1079, opened 2025-06-23
1056 featzen support grpc healthchecks service checks

### LINKED ISSUE #1056: feat(zen): support gRPC healthchecks / service checks
Like the other components in our system, we need to instrument the zen-consumer so that we can check on it via gRPC (mTLS) and poll the healthcheck endpoints from serviceradar-poller

mfreeman451:


mfreeman451 (commented):
lgtm

PR #1157, opened 2025-06-27
dont show navbar and sidebar on login screen

### LINKED ISSUE #1156: bug(ui): login page showing navbar/sidebar
When a user is presented with the login page that is all they should see, we're currently letting the navbar and sidebar bleed through on the login page to an unauthorized user.

mfreeman451:


mfreeman451 (commented):
lgtm

PR #1191, opened 2025-07-07
fixed snmp status in device inventory table

### LINKED ISSUE #1190: bug(ui): device inventory not showing SNMP icon
When the Device Inventory view loads, we briefly see the two devices that have SNMP collections flash the SNMP icon and then it disappears from the column/rows.

mfreeman451:


mfreeman451 (commented):
lgtm

PR #1196, opened 2025-07-07
flipping flag on compliance for armis checker

### LINKED ISSUE #1195: bug(sync): Armis updater incorrectly reporting availability
```golang
		if result, exists := resultMap[ip]; exists {
			attributes["SERVICERADAR_COMPLIANT"] = result.Available
		}
```

Needs to mark that field as non compliant (false) if host is availale

mfreeman451:


mfreeman451 (commented):
lgtm

PR #1215, opened 2025-07-08
1193 bugweb snmp dashboard doesnt support multiple collections

### LINKED ISSUE #1193: bug(web): SNMP dashboard doesnt support multiple collections
We have multiple sets of OIDs that we are polling (pairs of if/Out octet counters), it appears as if the 4 datasets are being merged into one graph.

mfreeman451:


mfreeman451 (commented):
lgtm

PR #1221, opened 2025-07-09
1220 bugpoller returning huge status messages in logs

### LINKED ISSUE #1220: bug(poller): returning huge status messages in logs
the poller is configured to poll most services on a 30s interval, in the case of the sync service, it returns a large amount of data and pollutes the logs, not sure if it is really a good idea to keep polling this thing every 30 seconds unless it is just for a healthcheck.

mfreeman451:


pkg/poller/poller.go
```diff
@@ -590,3 +656,54 @@ func (p *Poller) enhanceServicePayload(originalMessage, agentID, partition, serv
...
+		AgentId:     rp.agentName,
+		PollerId:    rp.pollerID,
+		Details:     rp.check.Details,
+	}
+
+	log.Printf("Sending GetResults request: %+v", req)
+
+	results, err := rp.client.GetResults(ctx, req)
+	if err != nil {
+		// Check if this is an "unimplemented" error, which means the service doesn't support GetResults
```
mfreeman451, line 675: this is brittle, need to improve this

mfreeman451: fixed

mfreeman451 (commented):
lgtm

PR #1247, opened 2025-07-12
1245 bugcore need to batch writes to proton

### LINKED ISSUE #1245: bug(core): need to batch writes to proton
we need to make sure we are batching all of our writes to proton, based on CPU utilization in a production deployment where proton is spiking, I think this is related.

mfreeman451:


mfreeman451 (commented):
lgtm

PR #1256, opened 2025-07-13
adding support to collect process info

### LINKED ISSUE #1255: feat(sysmon): collect process details
sysinfo already supports collecting information about processes, we should collect that and report on it

mfreeman451:


mfreeman451 (commented):
lgtm

PR #1271, opened 2025-07-14
1270 bugcore device registry n1 query

### LINKED ISSUE #1270: bug(core): device registry N+1 query
When processing devices in the device registry we are doing a proton query for every device seen, we could probably just make one query against the database for all devices, and then iterate through those.

mfreeman451:


mfreeman451 (commented):
lgtm

PR #1329, opened 2025-07-24
1327 featpoller send events to events stream

### LINKED ISSUE #1327: feat(poller): send events to events stream
In handlePoller, the logic for recovery (!ps.IsHealthy && !ps.LastSeen.Before(threshold) && ps.AlertSent) is a bit complex. This could potentially be simplified by having the ReportStatus function explicitly trigger a recovery event when it receives a healthy report from a poller that was previously marked as unhealthy.

mfreeman451:


mfreeman451 (commented):
lgtm

PR #1345, opened 2025-07-25
1306 featsysmon expose process metrics to UI

### LINKED ISSUE #1306: feat(sysmon): expose process metrics to UI
sysmon is collecting metrics from processes (pids) we need to update the core to store it and expose it via the API

- [ ] update core to support sysmon pid metrics
- [ ] update DB
- [ ] update SRQL
- [ ] update UI

mfreeman451:


mfreeman451 (commented):
lgtm

PR #1404, opened 2025-07-28
1346 featsync add blacklist for discovery

### LINKED ISSUE #1346: feat(sync): add blacklist for discovery
need to be able to configure networks in the config that we should be ignoring

mfreeman451:


mfreeman451 (commented):
lgtm

PR #1419, opened 2025-07-31
1348 featsync support config options for sweep settings even when using integration

### LINKED ISSUE #1348: feat(sync): support config options for sweep settings even when using integration

mfreeman451:


mfreeman451 (commented):
lgtm

PR #1426, opened 2025-07-31
1425 featlogger add otel style tracing support

### LINKED ISSUE #1425: feat(logger): add OTEL style tracing support
Need to inject trace and span IDs into standard and grpc-based request contexts so we can do tracing

mfreeman451:


mfreeman451 (commented):
lgtm

PR #1471, opened 2025-08-05
adding better logging and tests for multiple ASQ queries

### LINKED ISSUE #1467: bug(sync): multiple queries in armis integration over-writing sweep results
Need to verify that we're not creating a sweep.json after one ASQ query is processed, and then overwriting it after we run the second one.

mfreeman451:


mfreeman451 (commented):
lgtm

PR #1507, opened 2025-08-12
1487 featsync support tcp or icmp in sync

### LINKED ISSUE #1487: feat(sync): support tcp or icmp in sync
At the sync.json asq query level,

we need to be able to support looking at ICMP and/or TCP, so in certain ASQs we dont want to consider ICMP results to mark the device as available, only TCP.

mfreeman451:


mfreeman451 (commented):
lgtm

PR #1555, opened 2025-08-21
1548 featsync armis add support for multiple ipv4 addresses

### LINKED ISSUE #1548: feat(sync-armis): add support for multiple IPv4 addresses
Armis will return an array of IP addresses for a device. Right now we just pop the first one off the list and consider that the primary IP address. We now would like to take all of those IP addresses and associate it with a device. Not sure the best way to do this, but the first way I can think of is to just add it to metadata.

It starts to get murky if we wanted to consider these as additional datapoints when trying to merge / de-duplicate devices, so we probably shouldn't even bother with that right now.

- [x] Populate additional IP addresses in metadata
- [ ] Store additional metadata in KV
- [ ] Update agent/sweeper service to read metadata
- [ ] Ping/TCP scan additional hosts

~~It is especially unclear now after writing this out, how we are going to report on all of it without taking all of these additional IP addresses and creating new, unique device records.~~

~~We might just have to add all of the results from the ICMP or TCP checks to the metadata, and then when we run the armis updater and run our query `show devices ...` we process that metadata~~

If we add the results of the additional ICMP/TCP checks into the metadata for the primary device record, since Armis treats everything as just one ArmisID, we look at the results of all of those checks, if anything comes back as available, we update that one device record in Armis and mark that the device is not-compliant.

mfreeman451:


mfreeman451 (commented):
lgtm

PR #1582, opened 2025-08-28
1579 bugscan tcp syn scanner too fast

### LINKED ISSUE #1579: bug(scan): tcp syn scanner too fast
We're taking out the connection tracking table on routers because we're too fast, need to put in a knob to scale this back and add some docs on what you can do to mitigate this on a router and linux host.

mfreeman451:


mfreeman451 (commented):
lgtm

PR #1616, opened 2025-09-04
SRQL/OCaml POC

mfreeman451:
This is a PoC/MVP SRQL replacement written or rewritten in OCaml, from golang+antlr


mfreeman451 (commented):
lgtm

PR #2713, opened 2026-02-06
2293 feat rbac buildereditor

### LINKED ISSUE #2293: feat: RBAC builder/editor

We need a way to be able to create new RBAC profiles from the UI. This is something that only an admin or some kind of super_admin maybe (we would like to deprecate the concept of super_admin, but maybe it is worth keeping) could only build these things, the UI doesn't have to be super crazy but the main feature of this that we are looking for here is the ability to create groups of users and basically have an ACL that says what they can and cannot do on the system, beyond what we have today which is essentially `view` or `admin`.

Whatever we come up with has to be able to be evaluated fast, and it has to be idempotent -- live in the CNPG, loaded at startup, enforced at the API level (api lives in web-ng mostly anyways).

This thing doesn't really need to specifically be tenant aware since every tenant now gets their own schema in CNPG. We will also soon be moving to true tenant process isolation and every tenant will get their own core-elx and web-ng instance and whatever else.

**Additional context**

mfreeman451:

```
```

- [ ] I have signed the DCO?


elixir/serviceradar_core/lib/serviceradar/identity/changes/normalize_permission_keys.ex
```diff
@@ -0,0 +1,27 @@
+defmodule ServiceRadar.Identity.Changes.NormalizePermissionKeys do
+  @moduledoc """
```
mfreeman451: this whole thing needs to go

elixir/serviceradar_core/lib/serviceradar/identity/rbac/catalog.ex
```diff
@@ -0,0 +1,398 @@
...
+        "observability.events.view",
+        "observability.netflow.view",
+        "observability.alerts.view",
+        "observability.rules.view"
+      ]
+    }
+  end
+
+  @doc """
+  Normalizes permission keys by expanding deprecated aliases and removing
```
mfreeman451: why do we have deprecated aliases in here?

elixir/serviceradar_core/lib/serviceradar/monitoring/alert.ex
```diff
@@ -340,7 +340,7 @@ defmodule ServiceRadar.Monitoring.Alert do
       authorize_if actor_attribute_equals(:role, :admin)
 
       # Allow AshOban scheduler (no actor) to send notifications
-      authorize_if always()
+      authorize_if ServiceRadar.Policies.Checks.ActorIsNil
```
mfreeman451, line 343: i wish we didnt have to just rely on the actor being nil and assuming it is for oban, why cant oban have its own actor or the actor be known instead of nil?

web-ng/lib/serviceradar_web_ng_web/controllers/api/edge_controller.ex
```diff
@@ -83,7 +89,8 @@ defmodule ServiceRadarWebNG.Api.EdgeController do
       downstream_spiffe_id: params["downstream_spiffe_id"]
```
mfreeman451, line 89: edge agents dont use spiffe

web-ng/lib/serviceradar_web_ng_web/live/settings/auth_users_live.ex
```diff
@@ -278,15 +498,163 @@ defmodule ServiceRadarWebNGWeb.Settings.AuthUsersLive do
...
+        nil -> nil
+        profile -> profile.id
+      end
+    end
+  end
+
+  # Backwards-compat: accept either legacy `role` string or new `role_profile_id`.
+  # We always persist both `role` and `role_profile_id` so the app doesn't rely on
+  # base role for permissions, but legacy role checks won't accidentally escalate.
+  defp resolve_access_profile(profile_id, legacy_role, profiles) do
```
mfreeman451, line 580: need to get rid of these legacy profile checks

web-ng/test/serviceradar_web_ng/inventory_test.exs
```diff
@@ -25,8 +27,19 @@ defmodule ServiceRadarWebNG.InventoryTest do
       }
     ])
 
-    # Query using Ash through the Inventory context
-    devices = Inventory.list_devices(limit: 1_000)
```
mfreeman451, line 29: why did we get rid of this?

mfreeman451 (commented):
lgtm

PR #2806, opened 2026-02-12
wip: adding management_device_id

### LINKED ISSUE #2783: bug: unable to collect metrics from device using its public IP address
**Describe the bug**
https://demo.serviceradar.cloud/devices/sr%3A7588d12c-e8da-4b9e-a21d-8cc5c7faef38/interfaces/ifindex%3A4

`in:interfaces device_id:"sr:7588d12c-e8da-4b9e-a21d-8cc5c7faef38" interface_uid:"ifindex:4" latest:true limit:1`

We tried to setup SNMP polling on a couple of the interfaces on a device that was discovered via SNMP/discovery engine. DIRE (device identity and reconciliation engine) created a new device record for this, but we should have probably created some relationship between the parent device and this so we could reach it via SNMP polling, since we've created a device record based on this discovered public IP address, but it is unreachable across the network due to firewall that we are not opening up. 

I'm guessing we don't keep track of this right now and we probably should, one place to do this would just be in the apache age topology graph, and the poller would have to do a check first or maybe do some kind of fallback where if we can't reach the public IP address, we lookup the parent and query it instead. Apache age lookups are probably overkill and would be way slower than just adding this information to the device in `platform.ocsf_devices` schema

**To Reproduce**

**Expected behavior**

**Screenshots**

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Additional context**

mfreeman451:

```
```

- [ ] I have signed the DCO?

### DIFF
--- a/elixir/serviceradar_core/lib/serviceradar/agent_config/compilers/snmp_compiler.ex
+++ b/elixir/serviceradar_core/lib/serviceradar/agent_config/compilers/snmp_compiler.ex
@@ -460,8 +460,8 @@ defmodule ServiceRadar.AgentConfig.Compilers.SNMPCompiler do
       Logger.debug("SNMPCompiler: skipping device #{device.uid} (no OIDs)")
       nil
     else
-      # Get host address - prefer IP, fall back to hostname
-      host = device.ip || device.hostname
+      # If device has a management device, use its IP for polling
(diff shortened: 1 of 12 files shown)


mfreeman451 (commented):
lgtm

PR #2841, opened 2026-02-15
Features/netflow V10 bpg support

mikemiles-dev:
## Describe your changes
Extending Netflow Collector to Ingest V10 BGP Fields, Pipeline, Testing, and UI changes.

- [x] I have signed the DCO?


cmd/consumers/netflow-ingestor/main.go
```diff
@@ -0,0 +1,71 @@
+package main
```
mfreeman451: check out the Broadway (elixir) stuff there, there is an EventWriter.Producer that is sending messages to NATS JetStream should be a similiar pattern for Consumers
https://github.com/carverauto/serviceradar/blob/staging/packaging/zen/rules/netflow_to_ocsf.json

pkg/api/netflow_bgp_handler.go
```diff
@@ -0,0 +1,284 @@
+package api
```
mfreeman451: this seems like it was built around the old golang API, need to re-do this in web-ng where the new API is

pkg/consumers/netflow-ingestor/service.go
```diff
@@ -0,0 +1,179 @@
+package netflowingestor
```
mfreeman451: should be able to replace this with an elixir `Broadway` based consumer

pkg/db/cnpg_netflow.go
```diff
@@ -26,9 +26,11 @@ INSERT INTO netflow_metrics (
     sampler_address,
```
mfreeman451, line 26: I think all of these changes in `pkg/db` can be dropped and we just move this all into Elixir/Broadway and re-use existing Elixir models/schema

pkg/models/netflow.go
```diff
@@ -34,6 +34,8 @@ type NetflowMetric struct {
 	IPTos            uint32    `json:"ip_tos"`
 	VlanID           uint32    `json:"vlan_id"`
 	BgpNextHop       string    `json:"bgp_next_hop"`
+	AsPath           []uint32  `json:"as_path,omitempty"`           // BGP AS path sequence
```
mfreeman451: need to find the Elixir model and update that instead I think

rust/netflow-collector/src/converter.rs
```diff
@@ -640,6 +657,82 @@ fn field_value_to_protocol_name(value: &FieldValue) -> String {
...
+                    u32::from_be_bytes([chunk[0], chunk[1], chunk[2], chunk[3]])
+                })
+                .collect()
+        }
+        _ => vec![],
+    }
+}
+
 #[cfg(test)]
 mod tests {
```
mfreeman451, line 729: should these tests be in their own file?

go.sum
```diff
@@ -159,6 +159,8 @@ github.com/spiffe/go-spiffe/v2 v2.6.0/go.mod h1:gm2SeUoMZEtpnzPNs2Csc0D/gX33k1xI
 github.com/spiffe/spire-api-sdk v1.14.1 h1:pAVoElOJseU+L4ecrPFv8krIDG8XYWBYKztGUG5qWfg=
 github.com/spiffe/spire-api-sdk v1.14.1/go.mod h1:9hXJcMzatM1KwAtBDO3s6HccDCic++/5c2yOc5Iln8Y=
 github.com/stretchr/objx v0.1.0/go.mod h1:HFkY916IF+rwdDfMAkV7OtwuqBVzrE8GR6GFx+wExME=
+github.com/stretchr/objx v0.5.2 h1:xuMeJ0Sdp5ZMRXx/aWO6RZxdr3beISkG5/G/aIRr3pY=
```
mfreeman451, line 203: I'm guessing we can remove this as well

mfreeman451 (changes requested):
looks mostly ok, just need to re-do the consumer in Elixir and probably drop the `pkg/db` and API stuff.

rust/netflow-collector/src/converter.rs
mikemiles-dev: I can, not uncommon for rust to put tests on bottom of files as well.

mikemiles-dev: The #[cfg(test)] pattern is specifically designed for this use
  case. Rust developers expect to find unit tests at the bottom of the source file.

  The only reason to move them would be if:
  1. You prefer a cleaner source file (use option 2)
  2. You're writing integration tests that test the public API across modules (use option 3)

mfreeman451 (changes requested):
looks good, I think the only thing left to do is update the zen-consumer so it stops trying to process netflow in contention with the new Broadway consumer

elixir/serviceradar_core/lib/serviceradar/bgp/as_lookup.ex
```diff
@@ -0,0 +1,169 @@
...
+  Provides mapping of AS numbers to their registered organization names.
+  Results are cached to minimize external API calls.
+  """
+
+  use GenServer
+  require Logger
+
+  @cache_ttl :timer.hours(24)
+
+  # Well-known AS mappings (for common ASes, avoids API calls)
```
mfreeman451: i wonder if we could tie this into our ipinfo/geoip databases to lookup against their data instead of this table

elixir/serviceradar_core/lib/serviceradar/bgp/as_lookup.ex
```diff
@@ -0,0 +1,169 @@
...
+            Logger.debug("AS lookup failed for #{as_number}: #{inspect(reason)}")
+            {:reply, error, state}
+        end
+    end
+  end
+
+  ## Private Functions
+
+  # Fetch AS name from Team Cymru whois service
+  defp fetch_as_name(as_number) do
```
mfreeman451: do we need this if we already have an ASN lookup database from ipinfo?

elixir/serviceradar_core/lib/serviceradar/application.ex
```diff
@@ -74,6 +74,9 @@ defmodule ServiceRadar.Application do
         # RBAC permission cache (shared ETS, must start after PubSub)
         ServiceRadar.Identity.RBAC.Cache,
 
+        # AS Lookup cache for BGP routing (DNS-based whois with ETS caching)
+        as_lookup_child(),
```
mfreeman451, line 81: see previous comment about ASN lookups using ipinfo/geomax db

web-ng/assets/js/app.js
```diff
@@ -1869,6 +1869,95 @@ const Hooks = {
       }
     },
   },
+
+  /**
+   * Time Series Chart Hook for BGP traffic over time
+   */
+  TimeSeriesChart: {
```
mfreeman451: this might be a little ugly to merge in here, i would synchronize this branch against staging here (pull in latest changes) because app.js has been completely refactored, this stuff will just need to go into its own file/directory or whatever (we split app.js up and it is bare bones now as it should be)

elixir/web-ng/lib/serviceradar_web_ng_web/router.ex
```diff
@@ -345,6 +345,7 @@ defmodule ServiceRadarWebNGWeb.Router do
       live("/alerts/:alert_id", AlertLive.Show, :show)
       live("/observability", LogLive.Index, :index)
       live("/netflow", NetflowLive.Visualize, :index)
+      live("/bgp-routing", BGPLive.Index, :index)
```
mfreeman451: should this be a new view or should it live under observability?

mfreeman451 (commented):
check out comments please

mfreeman451 (commented):
lgtm

PR #2922, opened 2026-02-26
Fix flow-collector docker-compose mTLS config

mfreeman451:
## Summary
- Add `docker/compose/flow-collector.docker.json` with proper mTLS security and NATS creds file, matching the pattern used by trapd and other services
- Update `docker-compose.yml` to mount the docker-specific config (instead of the bare `rust/flow-collector/flow-collector.json`) and add the NATS creds volume mount

The flow-collector was failing to connect to NATS with `UnknownIssuer` because the config had `nats://` (no TLS) and was missing the NATS creds file. After this fix, `git pull && docker compose up -d` works cleanly.

## Test plan
- [x] `docker compose up -d` - all 14 containers up and healthy
- [x] flow-collector connects to NATS with mTLS: `Connected to NATS at tls://nats:4222 and ensured stream 'events' exists`
- [x] Both sflow (6343) and netflow (2055) listeners running

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -594,8 +594,9 @@ services:
       - "2055:2055/udp"       # NetFlow/IPFIX
       - "6343:6343/udp"       # sFlow
     volumes:
-      - ./rust/flow-collector/flow-collector.json:/etc/serviceradar/flow-collector.json:ro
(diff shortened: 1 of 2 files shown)


mfreeman451 (commented):
lgtm

PR #2923, opened 2026-02-26
Fix otel_trace_summaries schema: explicit platform prefix + repair migration

mfreeman451:
## Summary
- Fix migration `20260210120000` to use explicit `platform.` schema prefix instead of relying on `search_path`, which caused the table to land in `public` on some deployments
- Add idempotent repair migration `20260226230000` that handles all broken states:
  - Drops stale copies in `public` schema (table or materialized view)
  - Converts `platform.otel_trace_summaries` from materialized view to table if still a MV
  - Adds `refreshed_at` column if missing from existing table
  - Ensures indexes exist

**Root cause**: The `CREATE TABLE IF NOT EXISTS otel_trace_summaries` in the original migration had no schema prefix. Depending on the `search_path` at migration time, this could create the table in `public` instead of `platform`, which then triggered the startup migration's "public schema has ServiceRadar tables" guard.

## Test plan
- [x] Docker compose stack starts cleanly with all 14 containers healthy
- [x] `otel_trace_summaries` table exists in `platform` schema with all 15 columns including `refreshed_at`
- [x] web-ng `RefreshTraceSummariesWorker` runs without errors
- [x] Repair migration is idempotent — safe to run on already-fixed databases

### DIFF
--- a/elixir/serviceradar_core/priv/repo/migrations/20260210120000_convert_trace_summaries_to_table.exs
+++ b/elixir/serviceradar_core/priv/repo/migrations/20260210120000_convert_trace_summaries_to_table.exs
@@ -10,13 +10,20 @@ defmodule ServiceRadar.Repo.Migrations.ConvertTraceSummariesToTable do
   """
   use Ecto.Migration
 
+  @schema prefix() || "platform"
+
   def up do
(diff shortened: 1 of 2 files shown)


mfreeman451 (commented):
lgtm

PR #3148, opened 2026-05-03
feat(flow-collector): horizontal scale via NATS KV TemplateStore + Da…

mikemiles-dev:
…emonSet

Wires the new netflow_parser 1.0.3 TemplateStore extension point into the flow-collector and switches the K8s deployment from a single Deployment to a per-node DaemonSet behind a `Service: LoadBalancer` with `externalTrafficPolicy: Local`. Together these two changes let the flow collector scale horizontally with node count while preserving exporter source IP and sharing learned NetFlow / IPFIX templates across pods.

Why
---
A single flow-collector Deployment was a singleton bottleneck for UDP ingest. Scaling it required either source-IP-affinity routing (so every exporter always lands on the same pod and templates stay local) or a shared template store so any pod could decode any flow. We picked the shared-store path because it doesn't depend on the LB layer.

The original plan was to front this with Envoy Gateway UDPRoute, but a research spike found that Envoy Gateway's UDPRoute does not preserve client source IP (https://gateway.envoyproxy.io/docs/tasks/traffic/udp-routing/) — upstreams see the gateway pod's IP, which would collapse AutoScopedParser's per-(source_ip,source_id) scoping. Service+DaemonSet +`externalTrafficPolicy: Local` is the workable alternative on MetalLB (L2 mode preserves source IP through to the pod). Gateway API stays available for the rest of the stack (HTTP/gRPC/mTLS).

Code changes
------------
* `Cargo.toml`: bump netflow_parser 1.0.0 -> 1.0.3 (TemplateStore API); add `bytes` for the NATS KV payload type.

* `src/template_store.rs` (new): NatsKvTemplateStore implements the parser's TemplateStore trait. async_nats is async-only and the trait is sync, so we bridge via `tokio::task::block_in_place` + `Handle::block_on`. Safe under the multi_thread runtime the collector already uses. NATS KV keys are restricted to [A-Za-z0-9._=/-]; the scope produced by AutoScopedParser ("v9:1.2.3.4:2055/0", IPv6 with brackets, etc.) is sanitized via underscore replacement. Five render tests cover IPv4, IPv6, empty scope, distinct-scopes and distinct-kinds.

* `src/netflow/mod.rs`: NetflowHandler::new accepts an optional `Arc<dyn TemplateStore>` and threads it through `NetflowParserBuilder::with_template_store(...)`. Adds the new TemplateEvent::Restored arm to the existing event-callback (logs at info). Adds `update_template_store_metrics()` which iterates each per-source NetflowParser via `AutoScopedParser::{ipfix,v9,legacy}_info()`, sums `template_store_*` counters from each `CacheMetrics`, and writes the totals into ListenerMetrics atomics. Called under the parser lock inside parse_datagram so the snapshot is always fresh on the next scrape.

* `src/listener.rs`: build_handler now takes `Option<Arc<dyn TemplateStore>>`. sFlow ignores it (template-less).

* `src/config.rs`: new optional TemplateStoreConfig `{ kv_bucket, kv_history, kv_ttl_secs }`. Two deserialization tests.

* `src/main.rs`: new `bootstrap_template_store()` opens an *independent* NATS connection (separate from the publisher's), gets-or-creates the KV bucket idempotently, and wraps it in NatsKvTemplateStore. Kept on its own connection so KV-bucket flakiness can't backpressure publish and vice versa.

* `src/metrics.rs`: ListenerMetrics gains template_store_restored, template_store_codec_errors, template_store_backend_errors, source_count atomics. New `render_prometheus()` produces text exposition format (8 metrics, HELP/TYPE headers, protocol + listen_addr labels). New `run_prometheus_server()` is a hand-rolled HTTP/1.1 server on tokio's TcpListener — `GET /metrics` returns the exposition; anything else 404s. No new deps. Spawned from main.rs when `metrics_addr` is set; runs independently of the log reporter and the publisher so a scrape failure can never stall ingestion.

K8s manifest changes
--------------------
* `k8s/demo/base/serviceradar-flow-collector.yaml`: Deployment ->
  DaemonSet, replicas dropped, unused PVC removed (the Rust binary writes nothing to disk). ConfigMap JSON gains the `template_store` block. Service was already `externalTrafficPolicy: Local` — kept. Added `RollingUpdate` with `maxUnavailable: 1`.

* `helm/serviceradar/templates/flow-collector.yaml`: same kind flip, PVC removal, plus `nodeSelector`/`tolerations` knobs.

* `helm/serviceradar/values.yaml`: drop `replicaCount` and `data.*`, add `flowCollector.config.template_store` block, default `service.type: LoadBalancer` and `service.externalTrafficPolicy: Local`, add the IPFIX (4739) and metrics (50046) ports.

Tests
-----
48 tests pass:
  - 5 NatsKvTemplateStore key-render tests
  - 2 TemplateStoreConfig deserialization tests
  - 1 Prometheus exposition format test
  - 40 pre-existing flow-collector tests (unchanged) clippy --all-targets -D warnings: clean

What's left to do
-----------------
Code:
  - Prometheus scrape annotations on the Service so the existing Prometheus / ServiceMonitor picks up the new /metrics endpoint automatically. Both manifests need this.
  - PrometheusRule alerts: template_store_backend_errors > 0 sustained
    >5min = NATS unhealthy; template_store_codec_errors > 0 ever =
    wire-format mismatch (drain bucket).
  - Optional: TemplateStoreConfig.kv_replicas hint so the bucket is created with the right JetStream replication for HA. Currently defaults to 1 — operators can pre-create the bucket with `nats kv add flow_templates --replicas=3` to override.

Operational (no cluster access from here):
  - Run the source-IP smoke test against staging: LB_IP=$(kubectl get svc serviceradar-flow-collector \ -o jsonpath='{.status.loadBalancer.ingress[0].ip}') kubectl debug $POD -it --image=nicolaka/netshoot -- tcpdump -nn -i any port 2055 # send a packet to $LB_IP:2055 from a known IP # tcpdump should show that source IP, not MetalLB / node IP
  - Run the cross-pod template-share smoke test: send template+data, delete the receiving pod, send same data record, verify "Template restored from secondary store" log on the new pod and that NATS KV `nats kv ls flow_templates` shows scoped keys.
  - Decide JetStream stream replicas for HA on multi-AZ clusters.
  - Decide TTL on KV entries (default 0 = forever; 24h is reasonable if exporters re-announce frequently).

Decisions still open:
  - Whether to ship Prometheus scrape annotations + PrometheusRule in this branch or a follow-up.
  - Whether to expose `kv_replicas` in TemplateStoreConfig or rely on pre-create.

```
```

- [ ] I have signed the DCO?

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2290,6 +2290,17 @@ dependencies = [
 ]
 
 [[package]]
+name = "hashbrown"
+version = "0.17.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "4f467dd6dccf739c208452f8014c75c18bb8301b050ad1cfb27153803edb0f51"
+dependencies = [
+ "allocator-api2",
+ "equivalent",
+ "foldhash 0.2.0",
+]
+
+[[package]]
 name = "heck"
 version = "0.4.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -3111,11 +3122,11 @@ dependencies = [
 
 [[package]]
 name = "lru"
-version = "0.16.3"
+version = "0.18.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "a1dc47f592c06f33f8e3aea9591776ec7c9f9e4124778ff8a3c3b87159f7e593"
+checksum = "8a860605968fce16869fd239cf4237a82f3ac470723415db603b0e8b6c8d4fb9"
 dependencies = [
- "hashbrown 0.16.1",
+ "hashbrown 0.17.0",
 ]
 
 [[package]]
@@ -3383,9 +3394,9 @@ dependencies = [
 
 [[package]]
 name = "netflow_parser"
-version = "1.0.0"
+version = "1.0.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "c0e1f7859a3708e5f915bf00cc8e26dcef316a697d0f64fb01e017c054d2c677"
+checksum = "84bdb4541f8501a0fb36c546a617d5296c93875e0d9db406fe44f6966385c6e8"
 dependencies = [
  "lru",
  "nom",
@@ -5904,6 +5915,7 @@ version = "0.1.0"
 dependencies = [
  "anyhow",
  "async-nats",
+ "bytes",
  "clap 4.6.0",
  "env_logger 0.11.9",
  "flowparser-sflow",
--- a/docs/docs/flow-collector-scaling.md
+++ b/docs/docs/flow-collector-scaling.md
@@ -1,8 +1,8 @@
 ---
-title: Flow Collector — Scaling Guide
+title: Flow Collector -- Scaling Guide
 ---
 
-# Flow Collector — Scaling Guide
+# Flow Collector -- Scaling Guide
 
 Practical guidance for sizing the ServiceRadar flow collector, the knobs you can turn, what to watch in production, and when you've outgrown the default architecture. For protocol details (NetFlow v5/v9/IPFIX, sFlow), see the [NetFlow Ingest Guide](./netflow.md).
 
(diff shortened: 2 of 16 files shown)


mfreeman451:
Superseded by #4103, which is merged-ready with all checks green. Closing this
one — the template-store work landed, it just couldn't get there as a rebase.

**Why a new PR rather than updating this branch:** this PR predates the
Forgejo -> GitHub migration, which re-rooted the repository. This branch and
`staging` descend from different root commits, so they share no merge base at
all — `git merge-base` returns empty, and GitHub can only report
`CONFLICTING`, which reads like ordinary drift. #4103 re-derives the two
commits by cherry-picking them onto current `staging` (each commit diffed
against its own parent), which turned a whole-tree conflict into ~30
resolvable hunks.

**What was kept**, essentially as you wrote it:

* `template_store.rs` — `NatsKvTemplateStore` and the `block_in_place` bridge
  from async_nats to the parser's sync trait.
* `nats_client.rs` — the TLS/creds-aware connect helpers, now used by the
  template-store bootstrap.
* The Prometheus exposition and hand-rolled HTTP server, including the sFlow
  filtering for the structurally-zero `template_store_*` rows.
* The 1Hz metrics ticker with retired-source accounting, which is the right
  answer to the per-datagram O(sources) scan qodo flagged here.

**What was dropped:** the DaemonSet flip and PVC removal. That half rested on
"the Rust binary writes nothing to disk", which was true in May and isn't now
— `staging` since grew `ensure_owned_stream`, a JetStream stream
create/update reconciled from durable markers on a ReadWriteOnce PVC, with no
leader election. That's why the chart pins `strategy: Recreate`. A DaemonSet
would make N-concurrent-publishers the permanent steady state and can't share
a RWO PVC. Making stream ownership singleton-safe is the real prerequisite for
multi-pod ingest, and deserves its own issue rather than riding along here.

Also moot now: the `netflow_parser` bump — `staging` is already on 1.0.6.

**Two things only a live deploy surfaced**, both fixed in #4103:

1. The `TemplateStore` trait is synchronous and the parser calls get/put while
   holding the parser mutex, so each KV round-trip stalls all parsing for that
   listener. async-nats defaults to a 5s request timeout, far too long to hold
   the parse lock — now capped at 500ms and reported as a `Backend` error the
   parser already treats as a miss.

2. The flow-collector NATS user is granted `$JS.API.>`, which covers
   `create_or_update_key_value`, but KV get/put/delete ride on
   `$KV.<bucket>.>` directly and that was in neither the publish nor the
   subscribe allow list. The feature came up looking completely healthy —
   bucket created, ticker running, `/metrics` serving — and silently stored
   nothing.

Verified end-to-end on the farm01 cluster: after the permissions fix, a pod
restart restored template IDs 259/261/262 from KV — the same IDs that logged
`data lost` on the previous cold start — with
`flow_collector_template_store_restored_total` reading exactly 3 and zero
codec or backend errors.

Thanks for this — the design held up across four months of drift, and the
shared store is doing exactly what you built it to do.

PR #3340, opened 2026-07-13
fix(build): Fixed cargo build by moving netprobe/ebpf crate in EPF cr…

marvin-hansen:
## Describe your changes

Fixed cargo build by moving netprobe/ebpf crate in EPF crate to the excluded section of Cargo.toml; Bazel configuration remains unaffected.

- [x] I have signed the DCO?


marvin-hansen:
This PR fixes a cargo config issue that prevents a cargo build on MacOS. 
The patch is necessary so I can work on the Causal Engine via Cargo and thus bypassing fixing up Bazel on non-x86 platforms.

PR #3805, opened 2026-08-19
fix(web-ng): persist dashboard renderer preferences

mfreeman451:
`api.preferences.set` has been a write to nowhere.

The host advertises the `dashboard.preferences.write` capability, checks it, mutates the in-memory settings map, and pushes a `dashboard_preference_update` event — whose LiveView handler ignored its params and returned the socket unchanged:

```elixir
def handle_event("dashboard_preference_update", _params, socket) do
  {:noreply, socket}
end
```

The call succeeded, the renderer believed it, and the value was gone on reload.

## No new table needed

Renderer preferences are per user and per dashboard route, which is exactly the key `DashboardUserPreference` already carries: a unique identity on `(user_id, target_type, target_id)` and a free-form `metadata` map, with a package dashboard already addressed as `target_type: :package` plus the route slug.

So preferences land in the `metadata` of the row that already records whether the dashboard is a favorite.

## Two halves

- **The handler** merges the incoming key into that row and upserts it. It carries `favorite` and `is_default` through explicitly: both are listed in `upsert_fields`, so omitting them would reset the row's dashboard-level favorite to the create action's default on *every* preference write.
- **`host_payload/7`** merges the user's stored map into the instance settings the renderer reads, so preferences survive a reload. Instance settings are the seed for a user who has set nothing; the user's own values win.

The merge is fiddlier than it looks — instance settings may carry string or atom keys depending on how they were persisted, while the renderer only reads the string form — so it lives in its own module with tests rather than inline.

## Known limitation

`metadata` is read-modify-written rather than updated atomically, so two tabs writing different keys in the same instant can lose one. Acceptable for view preferences; a custom atomic jsonb action is the fix if it ever matters.

## Testing

- 15 new tests (13 unit + 2 doctests) pass under `SERVICERADAR_ALLOW_DB_FREE_TESTS=1`, tagged `:db_free` so they run in that suite.
- `mix compile` clean, no new warnings. `mix format` clean.
- **Not run:** the DB-backed path (the upsert and the reload round-trip). No Postgres on this machine and the web-ng suite skips itself without one — wants a CI run before merge.

## Context

Found while building a RIDS dashboard package that stores per-user favorites and layout choice through this API. That renderer currently mirrors every write to `localStorage` and tells the user when a value is local-only; once this lands, the hosted value wins automatically with no renderer change.


mfreeman451:
recheck

PR #3820, opened 2026-08-20
fix(helm): make wait-migrations prove the schema is current, not present

mfreeman451:
## Problem

web-ng's `wait-migrations` init container probed:

```sql
SELECT 1 FROM platform.schema_migrations LIMIT 1
```

That only asks whether the table exists, so it passes against any schema that was ever migrated — including one left months behind by a skipped migration hook. `helm upgrade --no-hooks` skips `serviceradar-core-migrations`, which is a `pre-upgrade` hook, so the gate meant to catch exactly that case waved it through.

Seen in a customer namespace on v1.4.39: the schema sat at high-water `20260713021000` with **73 migrations pending** while the init container cheerfully reported `Migrations detected.` web-ng then started, hit missing `platform.prefix_tag_snapshots` and `platform.authored_dashboard_versions`, never went Ready, and was initially diagnosed as an OOM. The previous release never referenced those tables, which is how five weeks of drift stayed invisible.

## Change

Compare the applied high-water mark against the newest migration the chart revision ships:

- `core.migrations.expectedVersion` carries that timestamp and reaches the init container as `MIGRATIONS_EXPECTED_VERSION`.
- **Postgres does the comparison.** `version` is a bigint and these are 14-digit timestamps, which is not something to hand to `test -ge` in whatever `/bin/sh` the tools image ships.
- A stale schema now **fails** the init container and names the skipped hook, instead of deferring the failure to a confusing runtime error.
- `expectedVersion: ""` restores the existence-only probe, for operators running against a deliberately pinned schema.

A hardcoded version is only worth something while it tracks reality, so `//helm/serviceradar:migrations_expected_version_test` pins it to the newest file in `elixir/serviceradar_core/priv/repo/migrations` and fails with the value to use. It follows the existing `py_test` pattern in that BUILD file and consumes the `//elixir/serviceradar_core:migrations` filegroup that already exists for `//rust/integration-db` — which does this same filename-vs-`schema_migrations` comparison for the test-database template. This is the chart-side counterpart.

## Verification

`helm lint` clean; chart renders. I extracted the rendered init container script and executed it against a stubbed `psql`:

| scenario | result |
| --- | --- |
| table missing | exit 1, `platform.schema_migrations unreadable` |
| **schema behind** (the reported case) | **exit 1**, `applied=20260713021000, expected=20260819190000`, names `--no-hooks` |
| schema current | exit 0 |
| version check disabled | exit 0 |

The new test passes, and I confirmed it is not vacuous by reverting `expectedVersion` to the stale value — it fails with the intended diagnostic.

## Reviewer notes

- **This is a deliberate behaviour change.** Any deployment currently running `--no-hooks` against a stale schema will now fail at init rather than starting. That is the point, but it will surface on other clusters, so it is worth knowing before this ships.
- **`bazel` is not installed on the machine this was authored on**, so `//helm/serviceradar:migrations_expected_version_test` has only been run as plain Python. CI is the first real Bazel run; the part worth a look is the runfiles path resolution (`TEST_SRCDIR`/`TEST_WORKSPACE` with a source-tree fallback), since the test reads across packages into `elixir/serviceradar_core`.
- Only `web.yaml` has this init container today; core runs migrations itself via `StartupMigrations`, so nothing else needed the same guard.


mfreeman451 (commented):
lgtm

PR #3829, opened 2026-08-21
fix(feeds): stop advisory feeds rewriting the whole corpus every run

mfreeman451:
## What happened

The `demo` namespace was failing liveness/readiness probes across every workload — web-ng, agent-gateway, core-elx, and pods that never touch Postgres at all (nats, gobgp, falcosidekick). CNPG reported `Cluster in healthy state`; disk was 29% full and the replication slot retained 17 MB. The primary was drowning in its own WAL.

```
checkpoints are occurring too frequently (10 seconds apart)
checkpoint complete: wrote 21588 buffers (16.5%) ... distance=721735 kB
```

Every backend in the database was parked on `LWLock:WALWrite` / `IO:WalSync`.

## Root cause

The loader had **no skip guard**, so every `nist-nvd2` run re-upserted the entire corpus — ~360k advisories and ~2.5M coordinates — even though almost none had changed.

From `pg_stat_statements` on demo:

| statement | calls | WAL | per call |
|---|---|---|---|
| `INSERT advisory_coordinates` | 59,439 | **2,333 GB** | 42 MB |
| `INSERT vulnerability_advisories` | 22,120 | **3,193 GB** | 178 MB |
| `UPDATE ... SET current` (finalize) | 14,684 | **337 GB** | 25 MB |

~**5.9 TB of WAL** — more than the entire flow-attribution pipeline. From `pg_stat_user_tables`:

- `advisory_coordinates`: 2,557,231 live rows, **155,595,694** `n_tup_upd`, **`n_tup_hot_upd = 0`**
- `vulnerability_advisories`: 387,928 live rows, **111,057,647** `n_tup_upd`

At ~72 MB/s against `max_wal_size = 1 GB`, CNPG checkpointed every 10 s. That timed out the feed's own statements (`57014 query_canceled`), Oban retried, and the retry regenerated the storm. `nist-nvd2` had not completed successfully in three days.

## Changes

**Skip unchanged advisories.** `existing_modified_at/2` reads through `type(a.modified_at, :utc_datetime_usec)`. It is a schemaless query over a `timestamp without time zone` column, so Ecto types the field as `:any` and Postgrex returns a raw `%NaiveDateTime{}`; comparing that against the parser's `%DateTime{}` answers "changed" for every record. Same hazard and same fix as the existing precedent in `topology_graph/canonical_rebuild.ex:139-147`. `same_modified?/2` normalizes both sides and uses `DateTime.compare/2`, never `==`, because the round-trip legitimately changes precision metadata (`{123000, 3}` vs `{123000, 6}`).

**`finalize/4` takes a required `:demote_missing`.** This is the sharp edge. Skipped rows keep an older generation while staying live, so the old blanket *"demote everything whose generation != N"* would hide the bulk of the corpus from the matcher and then hand it to `reap_old_generations/3`. That function now also requires `current == false` — both `advisory_coordinates` and `endpoint_vulnerability_matches` cascade off the advisory row. `Loader.full_sweep?/1` is the only sanctioned way to authorise a demote.

**Drop `:generation` from the coordinate `on_conflict` replace list.** That column is indexed, which forced every upsert to be a non-HOT update rewriting all six indexes including the 416 MB GIN trigram. `n_tup_hot_upd = 0` against 155M updates is that one field.

**Coordinate batch 2,000 → 500.** The old comment sized it against Postgrex's 65,535 bind-param cap, which is not the binding constraint: 2,000 rows is ~34k params, fits fine, and still produced 42 MB of WAL per statement — and with `log_min_duration_statement = 500ms`, made Postgres echo ~742 KB of bind parameters into its own log for every batch.

**`flush_chunk/5` is now transactional**, so a chunk that fails halfway cannot leave advisories pointing at a partial coordinate set.

**Observability.** `feed_worker` logs `advisories_skipped` and raises an error-level alarm when a run skips nothing against a non-empty stored corpus. An inert guard is otherwise indistinguishable from a healthy run — which is how this shipped.

## Tests

- Unit tests pin the `%NaiveDateTime{}` shape specifically. The `~U[...]` variants pass *even against the broken code*, which is exactly the blind spot that let this through.
- A new DB-backed `@moduletag :integration` module round-trips through real Postgres, because that boundary is where the bug lives and a hand-built map cannot reproduce it.

**Verified non-vacuous:** reintroducing the original defect fails 6 of the 7 integration tests, including the direct `%DateTime{}` vs `~N[2024-01-15 12:34:56.123000]` mismatch.

```
7 tests, 0 failures   (integration, real CNPG)
4560 tests, 1 failure (full DB-free unit suite; the one failure is
                       test/db/migrate_db_test.exs, which the Bazel
                       unit_tests target excludes and which needs a
                       provisioned database)
```
Credo `--strict` clean project-wide; `mix format --check-formatted` clean; `mix compile --warnings-as-errors` clean.

## Operational note

`nist-nvd2` is currently **disabled on demo** (`vulnerability_feed_definitions.enabled = false`, was `enabled=t, refresh=21600`) and the in-flight Oban job was cancelled, to stop the bleeding. Re-enable after this merges and deploys. No data was lost: 379,161 advisories remain `current = true` and visible to the matcher.

## Out of scope

- `max_wal_size = 1 GB` is undersized for this cluster — checkpoints ran every ~14 s even with the feed stopped. That is a CNPG/gitops tuning change, not a code change.
- The KEV parser emits `modified_at: nil`, so KEV will keep writing every run. Deliberate: `nil` means *unknown*, never *unchanged*. Low WAL impact.
- Two sibling instances of the same schemaless-query type hazard exist elsewhere in `serviceradar_core`; they deserve their own PR rather than being bundled here.


mfreeman451 (commented):
lgtm

PR #3855, opened 2026-08-22
docs(openspec): propose running the benchmarks that already exist

mfreeman451:
## Why

ServiceRadar has benchmarks in **all three runtimes and nothing executes any of them.**

| runtime | benchmarks | what runs them |
|---|---|---|
| Go | ~10 files, incl. `go/pkg/edge/edgerecord/sweep_ingress_benchmark_test.go` | nothing |
| Elixir | `serviceradar_core/bench/` — `sweep_ingress`, `event_writer_pull_buffering`, `metric_fixture_cnpg_insert`, `metric_fixture_profile` | nothing |
| Rust | `rust/netprobe/benches` | nothing |

No `-bench`, `benchmem`, `mix bench`, or `criterion` invocation exists in `.github/workflows/`, `//buildbuddy.yaml`, or the `Makefile`, and no Bazel target runs one. Go benchmarks are **compiled** by `bazel test` and skipped without `-bench`, so today they prove only that they still build. The Elixir `bench/*.exs` are scripts nobody invokes.

A benchmark nobody runs still passes review and still looks like coverage on a file listing. One is already load-bearing for something else: `bench/sweep_ingress_fixtures.exs` is the shared fixture matrix `sweep_bench_fixture_test.exs` requires at module level, and a missing Bazel input for it aborted an entire integration shard earlier this week.

## Two deliverables, not one

**Part 1 — run what exists.** Buildable now. Bazel targets that execute the benchmarks, results recorded as artifacts, and **no threshold gate**: a benchmark on a shared runner is noisy, and a threshold set before its variance is measured produces failures that are not regressions. The predictable response is to mute the check — and a muted gate still looks like protection.

**Part 2 — pipeline throughput.** Deliberately **blocked** on `unify-sweep-results-proto`'s first green vertical slice. No current benchmark measures throughput, and the one edge benchmark disclaims it in its own header:

> THIS IS NOT PRODUCTION INGRESS, and its numbers are not a capacity figure... There is NO extraction, NO `ValidateSweepRecord`, NO trust resolution, NO signature verification and NO decompression.

Built before the composed path exists, a throughput benchmark would stub most of it and measure its own stubs — the same defect that header already documents.

A real NATS JetStream in `sr-testing` is recorded as available, so neither the slice nor the throughput bench needs to mock the PubAck hop.

## Scope

- New capability `edge-pipeline-benchmarks`; `openspec validate --strict` passes.
- **Proposal only — no implementation**, per the OpenSpec approval gate.
- Not in scope: any threshold, any optimisation, any change to what the benchmarks measure.

## Risk, stated

This change is itself a horizontal proof surface, which the vertical-slice milestone warns against. Part 1 is small and protects work already paid for; Part 2 is the expensive half and is gated behind the composed result rather than allowed to precede it.


mfreeman451 (commented):
lgtm

PR #3856, opened 2026-08-22
fix: untrack the serviceradar_core deps symlink

mfreeman451:
`elixir/serviceradar_core/deps` is a **symlink tracked in git** on `staging` as of 038d2c24f2. My mistake — amending #3853 I ran `git add -A elixir/serviceradar_core`, which swept it up.

## Why it matters

The symlink stores an absolute path, so every other checkout gets a dangling link to `/Users/mfreeman/src/serviceradar/elixir/serviceradar_core/deps`. And in the primary clone — the very path it names — it resolves to **itself**, so checking out `staging` there replaces the real `deps` directory with a loop.

## Why .gitignore didn't stop it

Both existing rules end in a slash:

```
deps/
elixir/**/deps/
```

A trailing slash matches **directories only**. A symlink isn't a directory to git, so the ignore never applied and `git add -A` was free to stage it. Borrowing the primary clone's `deps` into an extra worktree is the normal workflow here, which makes this a trap anyone can hit.

Dropping the slash makes the rule match a file, directory, or symlink of that name. Same change for `_build`, which is borrowed the same way and has the same hole.

## Verification

Reproduced the mistake rather than reasoning about it:

- **Before:** create a `deps` symlink, `git add -A` → tracked.
- **After:** same sequence → stages nothing.

No build output or dependency content is affected; `mix deps.get` restores a real directory.


mfreeman451:
BazelCI failed twice on this branch before going green. Recording the evidence so it isn't re-litigated: **it's a flake in the Elixir integration shards, not this change.**

| run | SHA | result |
|---|---|---|
| 1 | `ae11c4559c` | `integration_tests_s2` FAILED |
| 2 | `57141fa8ad` | `integration_tests_s0` FAILED |
| 3 | `5b392b2aad` | **all pass** |

The three SHAs are byte-identical in content — runs 2 and 3 were bare `--amend` re-triggers with no edits, so only the committer date differs.

Three independent reasons it can't be this diff:

1. **A different shard fails each time, then none do.** Deterministic breakage doesn't move between shards.
2. **The change cannot affect test selection.** Shard membership comes from `ALL_TEST_SRCS`, which globs `test/**/*_test.exs` only. This PR touches `.gitignore` and removes a symlink at the package root — neither is under `test/`, so no test can move between shards.
3. **The failing shard is the slowest in its run each time** (125.7s vs 55–100s for its peers), which points at a timeout or resource contention rather than an assertion.

I could not retrieve the actual ExUnit failure: `bb view` doesn't inline `test.log`, so the assertion text isn't visible from the invocation. That's the one thing I'd still want in order to file it properly against the integration-DB flakiness already tracked for these shards.

PR #3861, opened 2026-08-22
fix(feeds): reclaim feed jobs whose node restarted in place

mfreeman451:
The orphan reclaim I added in #3853 misses the most common way a feed job's owner dies. I asserted there that "a replaced pod never reuses a node name" and tested only node liveness. True when the **pod** is replaced; false when the **container** restarts inside the pod — the pod keeps its IP, and since nodes here are named after the pod IP, the BEAM returns under the identical name.

## Caught in production within the hour

Deployed #3853 to farm01 and watched this happen:

```
06:04:00  nist-nvd2 attempt 1 starts on serviceradar_core@10.42.51.161
06:08:15  container OOMKilled (exit 137)
06:08:16  container restarts -- same pod, same IP, same node name
07:31     job still `executing`, 87 minutes in
```

The reclaim saw a live node, declined to act, and the row was on its way to the full **240-minute** Lifeline window this was written to avoid. The exact failure it was meant to prevent, one restart mode over.

## The fix

Presence isn't enough — it has to be the *same instance*. A node whose VM started **after** the job's `attempted_at` cannot be running that job, whatever it's called.

| condition | verdict |
|---|---|
| node absent from cluster | orphaned (pod replaced — already handled) |
| node present, VM started after `attempted_at` | **orphaned (new)** — restart in place |
| node present, start time unreadable | healthy (conservative) |
| no `attempted_at` | left to Lifeline |

This keeps the property that made the original approach worth having: no threshold on job age, and no change to the global `Lifeline` setting every other worker shares.

Start time comes from the monotonic clock via `:erlang.system_info(:start_time)`, **not** `:erlang.statistics(:wall_clock)` — the latter resets a global counter as a side effect of being read, and this runs on a path whose whole purpose is to disturb nothing.

## Verification

`89 tests, 0 failures` in the advisory_feeds suite; format, `--warnings-as-errors`, `credo --strict` clean.

Non-vacuous: restoring the liveness-only predicate fails **exactly** the restart test and nothing else.

```
  1) test orphan reclaim a job is orphaned when its node restarted after the attempt began
16 tests, 1 failure
```

## Not fixed here

**The OOMKill itself.** farm01 caps `core-elx` at 8Gi where demo allows 16Gi, and the NVD run doesn't fit — which is the real reason farm01's `nist-nvd2` hasn't succeeded since 2026-08-17. That's a capacity question about the feed's memory profile, not a scheduling one, and it wants its own change.

Refs #3853.


mfreeman451:
Reproduced a second time on farm01, on a different pod, after cancelling the first stuck job by hand. This is deterministic, not a one-off:

| run | job | started | node | OOMKilled | into the run | then |
|---|---|---|---|---|---|---|
| 1 | 974575 | 06:04:00 | `serviceradar_core@10.42.51.161` | 06:08:15 | 4m15s | stuck 103 min, cancelled manually |
| 2 | 981308 | 07:47:05 | `serviceradar_core@10.42.51.155` | 07:50:56 | 3m51s | stuck 50 min and counting |

Both exit 137, both on pods that kept their IP across the restart, so both came back under the **identical node name**. The liveness-only predicate returns `false` in both cases and the row waits out the 240-minute Lifeline window.

Worth noting run 2 is the cleaner evidence: it started on a pod with `restarts=0`, so there is no question of the node having been odd beforehand. `attempted_at` 07:47:05 against a VM that booted 07:50:56 — the comparison this PR adds — is unambiguous.

The OOM itself is still out of scope here (farm01 caps `core-elx` at 8Gi where demo allows 16Gi, and idle sits ~2.3Gi, so the NVD run spikes roughly 6Gi). Filed separately. This PR only stops the corpse from blocking the queue for four hours afterwards.

PR #3921, opened 2026-08-23
ci(elixir): move slow Mix quality checks off every PR

mfreeman451:
## Why

Elixir Quality was a required PR check that compiled each Mix project's full Hex closure (and Rustler NIFs for five of seven apps). That was 18m+ of wall time for little extra signal over `mix format` and Credo.

**Compile is still a PR gate.** BuildBuddy BazelCI already does `bazel build //...` and `bazel test //elixir/...` on every PR. If Elixir does not compile, that job is red. The Mix quality compile was a second, slower, non-hermetic compile of the same tree.

What Mix quality was adding on top of Bazel was `--warnings-as-errors`, xref, hex/deps audit, Sobelow, and the OpenAPI dump. Those stay, they just run daily instead of blocking merge.

## What PRs keep

`.github/workflows/elixir-quality.yml` still uses the same job names (`Elixir Quality (web-ng)`, etc.) so required checks keep matching:

- `mix format --check-formatted`
- `mix credo --strict`
- core proto drift (`make verify-proto-elixir`)
- DB baseline / migration safety

No rustup. `--lint-only` sets `SERVICERADAR_SKIP_NIF_COMPILATION=1`.

## What moves to a daily BuildBuddy cron

New action **Elixir Quality (daily)** in `buildbuddy.yaml`:

- `0 7 * * *` UTC against `staging`
- `./scripts/elixir_quality.sh --all --skip-dialyzer --skip-nif`
- then `mix serviceradar.openapi.dump --check`

`--skip-nif` is because the workflow-runner image has no gcc/cargo (by design). BazelCI already builds the NIFs on every PR.

The workflows pool is one replica. This cron and BazelCI queue on the same runner; 07:00 UTC is meant to miss US-hours PR traffic.

`elixir-dependency-audit.yml` remains the calendar advisory scan. PRs no longer duplicate `mix hex.audit` / `mix deps.audit`.

## Caveat

`elixir/palisade` is not a Bazel target. Its Mix compile is only on the daily job. It is a small app; format + Credo still run on PRs.


mfreeman451 (commented):
lgtm

PR #3932, opened 2026-08-23
feat(edge): ResolvedPrefix contiguous resolved-prefix tracker (task 3.3(c) part 1)

mfreeman451:
Task 3.3(c) part 1 of three. Pure data and functions — **no process**, which is not justified until the bounded asynchronous publisher arrives (parts 2 and 3).

Head `b3817dcac5`, based on `usp-01-proposal` including #3925. **Please review this SHA and request changes rather than merging on comments** — adopting the stricter rule from this PR onward.

## What it is

The Elixir port of the gateway-side contiguous resolved-prefix tracker. Frames publish asynchronously and earn durable outcomes out of order; this advances the resolved watermark across a contiguous run of RESOLVING dispositions, so the gateway reports a prefix the agent may act on without claiming anything about sequences behind a gap.

Semantics preserved from the Go `gwprefix` docs:

- **Four ABI kinds resolve; `REJECTED_RETRYABLE` never does** and caps the prefix exactly like a missing outcome. Advancing past it would eventually authorize reclaiming customer data that was never delivered.
- **But retryable is provisional, not a verdict** — a resolving kind supersedes it and advances the prefix, including across sequences queued behind it. Immutability there would wedge the lane forever, since the retry could never be recorded.
- **Two watermarks stay separate.** A PubAck resolves; only the agent's own durable terminal action reclaims, and one local event never vouches for an earlier sequence.
- **u64-max terminates** — membership-driven, not a `s <= seq` counter that wraps (the Go comment records that hanging).

**The Go `gwprefix` is retained deliberately** as a comparison oracle until this port is integrated and verified. Deleting it is a focused follow-up, not part of this PR.

## The transition table is derived, not listed

Every declared kind × every declared kind, policy stated once as data:

```
prior == new                           -> idempotent
prior == retryable and resolving?(new) -> allowed (supersedes)
otherwise                              -> conflict
```

Separate matrices for PENDING and RESOLVED sequences, because the policy differs — inside the prefix a resolving kind is terminal and nothing supersedes it.

## One assertion deleted for being vacuous

The review asked for whole-tracker comparison before and after every **rejected** transition. **In Elixir that cannot fail**: `{:error, reason}` exposes no replacement tracker, and a function cannot mutate the caller's term. It read as rigorous and proved nothing, so it's gone rather than left to overstate the suite.

The two properties that do bite, both mutation-confirmed:

| property | mutant that killed it |
|---|---|
| conflicts return an error, not `{:ok, corrupted}` | conflicts made last-write-wins |
| idempotent transitions return a tracker **equal** to the input | idempotent branch drops the pending entry |

"State unchanged after rejection" becomes observable once a process owns the tracker — that belongs to the pipelining stage.

## An ordering claim removed rather than defended

A comment claimed unknown kinds are refused *before* anything is written. `record/3` documents **which** errors it returns, not which wins when an input is invalid several ways at once — so reordering the guards is behaviour-preserving and a mutation doing so **survives, correctly**. Pinning a precedence would invent a promise the function doesn't make and would fail the next person who reorders guards for readability. The docstring now states precedence is unspecified; the test file records why that mutation survives by design.

## Verification

**Nine mutants killed** across two rounds: retryable made to resolve · the two watermarks conflated · one terminal event vouching for earlier sequences · the prefix advancing across a gap · retained evidence never released · undeclared kinds accepted · conflicts as last-write-wins · idempotent dropping state · (plus the earlier set). **One survives by design** and is documented as such.

Also pinned: watermark monotonicity across an interleaved run (neither moves backwards, reclaim never passes resolution, with a non-vacuity check), and terminal evidence arriving before its PubAck explicitly refused rather than retained.

757 tests, 0 failures. Credo-clean, format-clean.

## Next

- 3.3(c) part 2 — pipelined async publishes under bounded frame/byte/PubAck-deadline windows
- 3.3(c) part 3 — separately bounded bulk/interactive/recovery pools
- then `EdgeRecordIngestService.Stream`, then the first database vertical slice


mfreeman451 (commented):
**BLOCKING REVIEW** of exact head `b3817dcac53c1b9814ff1d7c09cb64662fa15104`.

GitHub refused `REQUEST_CHANGES` because the authenticated account is also the PR author. Treat this commented review as changes requested; a separate reviewer identity is required to enforce the state mechanically.

### P1 — the gateway tracker can never retire its per-frame state

`ResolvedPrefix` is explicitly gateway-side, but every contiguous result moves into `disposition` and the only deletion path is `record_terminal_outcome/2` (`resolved_prefix.ex:160-186,228-266`). That event is an **agent-local durable action**. The client→gateway protocol carries only `lane_open` and `delivery_frame`; it never reports that action. The approved design assigns local reclaim solely to the agent (`design.md:258-263`). A real gateway therefore retains one disposition per resolved frame for the lane lifetime. Calling `record_terminal_outcome/2` after sending an ACK would instead conflate transport send with agent durability.

Split the bounded gateway remote-prefix/ACK-window state from the Go agent's local reclaim state, or define a gateway retirement event that exists in the protocol. Remote PubAck tracking is 3.3(c); the two-runtime/two-watermark binding is task 3.10 and explicitly requires Go/Elixir shared vectors.

### P1 — impossible sequence values can wedge the tracker

`new/1` and `record/3` accept arbitrary Elixir integers although the wire and Go oracle are `uint64`. Reproduced: construct at `max + 1`, record authoritative `max + 1`, and `resolved` becomes `18446744073709551616`; terminal recording returns `{:ok, ...}` but reclamation remains at max and retains the impossible entry forever. Reject values outside the uint64 lane domain.

The constructor is also integration-ambiguous: `EdgeRecordLaneOpen` has `sequence_base` (MUST be 1) and `first_unresolved_sequence`. Since the constructor seeds `resolved = arg - 1`, its argument must be named/documented/bound as `first_unresolved_sequence`. Passing `sequence_base` on reconnect at first-unresolved 100 wedges on already-reclaimed 1..99.

### P1 — two load-bearing transition paths survive

- The retryable→resolving matrix puts seq 2 behind missing seq 1 and checks only `pending_disposition == new`. A mutant advancing after retry only for authoritative, leaving audit/quarantine/permanent pending, passes all 23 tests. Exercise every resolving kind at the leading gap and assert unwedging.
- Duplicate terminal evidence lacks whole returned-state equality. A mutant removing an already-queued terminal marker passes all 23 tests. Cover queued-behind-gap and already-reclaimed idempotence, using `===` rather than `==` for exact tracker equality.

N1 and N2 are valid kills; each compiled and died on exactly one assertion. N3 is correctly behavior-preserving.

### P2 — the transition matrix omits reclaimed sequences

After authoritative seq 1 is reclaimed, recording permanent or retryable for seq 1 returns `{:ok, unchanged}` because the original evidence was deleted (`record_inside_prefix/3`, lines 124-130). This may be a deliberate bounded-history no-op, but it is a third lifecycle policy and contradicts the “full matrix / terminal conflicts” claim. Define and test it explicitly.

### P2 — exact-head prose still contradicts the PR

- The moduledoc still says delete the Go oracle (`resolved_prefix.ex:14-19`).
- “Can be rebuilt from durable stream/DLQ” contradicts the approved restart/replay model and cannot reconstruct agent-local reclaim.
- The test still says every rejection asserts whole-state equality (`resolved_prefix_test.exs:135-141`) and repeats that in names/comments at 171, 201, 224, and 239-252; these reread the original immutable term.
- Reconcile task 3.3 as partially landed without checking it, and do not imply task 3.10 is complete without its shared vectors.

Verification: all GitHub checks green; uncached exact-head Bazel edge shard passed; `git diff --check` clean; no Bazel/runfiles defect.

mfreeman451:
Additional exact-head evidence for the P1 transition-suite finding:

- Mutating only retryable → `ACCEPTED_AUDIT_ONLY` to replace `pending` correctly **and also set `reclaimable = seq`** leaves all 23 tests green. In the matrix case that produces `reclaimable = 2` while `resolved = 0`—a direct deletion-authority violation. The allowed branch must assert the complete expected tracker/watermarks, not only `pending_disposition`.
- On a second terminal event already queued for seq 2 behind missing local terminal seq 1, a mutant setting `reclaimable = 2` also leaves all 23 tests green. The current duplicate test covers only an already-reclaimed seq 1. Add exact returned-state equality for the queued duplicate, then close seq 1 and prove the run advances normally.

Bookkeeping: the branch records six first-round kills plus N1/N2 killed and N3 survived: **8 killed / 1 behavior-preserving survivor**, not 9/1, unless a ninth killed mutation is enumerated.

PR #3937, opened 2026-08-23
fix(topology): suppress redundant inferred-segment edges, iterate satellite placement

mfreeman451:
## Summary
Follow-up to #3928 after demo/farm01 rollout showed two regressions in the first cut:

1. **Redundant doubled edges / cluster inflation.** Device-to-Device inferred-segment links carry `relation_type: ATTACHED_TO`, and evidence normalization maps that to `endpoint-attachment` before anything else sees the raw value. The new read-model rows therefore masqueraded as endpoint-attachment edges: demo attachment count jumped 24 -> 133, endpoint clusters absorbed duplicate membership, and every cluster path rendered twice as long crossing lines.
   Fix: filter raw **links** before edge construction - drop an inferred-segment link when *both* endpoints already appear in another link. Survivors keep otherwise-isolated devices connected (the original anti-island goal) and flow through normal attachment clustering under their anchor.

2. **Satellite placement not transitive.** `attachmentSatellitePlacements` only attached satellites to already-placed anchors, so a residual switch (no backbone edge) and all of its endpoints fell into the unplaced 8-row lanes (farm01: two neat node columns with edges sweeping the whole canvas).
   Fix: iterate placement to a fixpoint (max 4 rounds) so chains resolve through newly placed satellites - a residual switch rings around its placed router, then its endpoints ring around the switch.

Also removes the unreachable `inferred-segment` clause in `edge_topology_class` (evidence normalization maps it to `inferred` before the class lookup).

## Test plan
- [x] New regression test `latest_snapshot/0 keeps inferred-segment edges only for otherwise-isolated devices` (god_view_stream_test.exs)
- [x] New JS chain-placement test; god_view suite 199/199, eslint + tsc clean, contract tests green
- [x] Full `god_view_stream_test.exs` + channel tests: 102 tests, failures identical to the 43 pre-existing on staging base (verified by base run)
- [ ] farm01 demo-cluster visual verification (in progress)

Refs #3928. OpenSpec: `fix-topology-islands-and-cluster-expansion`.


mfreeman451:
Superseded by #3999, which includes this regression fix and replaces the remaining client-side topology placement/routing with the fully tested ELK-authoritative scene implementation.

PR #3938, opened 2026-08-23
fix(edge): the gateway holds gateway state; reclaim belongs to the agent

mfreeman451:
Corrections to `b3817dcac5`, which merged as #3932 before these were ready. **Please review this head and request changes rather than merging on comments.**

## P1 — the gateway could never call `record_terminal_outcome/2`

You were right that this is a design error, not a missing feature. Local durability is the **agent's** fact; the gateway cannot observe it. The consequence you traced is exactly the trap: retained dispositions grow forever unless the gateway equates *sending* an ack with the agent having durably acted — the precise false equivalence the two-watermark split existed to prevent.

Reclaim now lives with the Go agent. This module keeps only what the gateway can see: base, the resolved watermark, pending outcomes, retained dispositions.

Release is driven by `EdgeRecordLaneOpen.first_unresolved_sequence`, which the agent reports on open and resume — an **observation**, not an inference. A structural test asserts no gateway-side API can advance it (`record_terminal_outcome/2` and `reclaimable_through/1` no longer exist).

## P2 — the post-release hole is closed structurally

A released sequence had no retained disposition, so a contradicting outcome found nothing to contradict and returned `{:ok, t}` — silence reading as agreement. Release now advances `base`, so those sequences are out of range and fail `:below_base`. The `:error -> {:ok, t}` branch that created the hole is **gone**, not special-cased.

## P1 — lane bounds are the protobuf uint64 range

A sequence past uint64 max was accepted and advanced the watermark to a value nothing downstream can express. Now `:above_lane_max`.

The constructor takes **`first_unresolved_sequence`**, documented against `sequence_base`: the two coincide only on a fresh lane, since `sequence_base` MUST be 1 and names the origin while `first_unresolved_sequence` is where the agent still needs work. Seeding a resumed lane from the origin re-opens a window the agent already closed.

## P1 — whole state, with `===`

Transitions assert the complete returned tracker **and** both observable watermarks using `===`, so a float in a watermark cannot pass and a supersession cannot move the prefix while only the pending entry is checked. Both mutants you reproduced die.

## P2 — stale claims corrected

Task ownership is **3.5** — that is where the exhaustive resolving-PubAck list lives, including the security-quarantine one that collapses onto `ACCEPTED_QUARANTINE` on the wire; 3.3(c) consumes it. The Go `gwprefix` guidance now says **retained as a comparison oracle** until this port is integrated.

## Mutation accounting, corrected

You were right that the count was wrong, and the cause was summing rounds rather than counting what still applies.

**Nine killed** against the current implementation: retryable made to resolve · prefix advancing across a gap · undeclared kinds accepted · conflicts as last-write-wins · idempotent branch dropping state · supersession advancing past a missing sequence · a sequence past uint64 max · release not advancing base · release inferred from resolution rather than the agent's report.

**One survives by design**: reordering the validation guards is behaviour-preserving against a contract that documents which errors exist, not their precedence.

**Three are retired with the code they targeted** — "watermarks conflated", "one terminal event vouches for earlier sequences", and "retained evidence never released" all tested the reclaim path this revision removes. Carrying them forward is how the number inflated.

## Verification

Whole-repository sweep green: **192 tests pass**. Edge shard uncached green, credo-clean, format-clean.

## Process note

Branch protection is now configured on `usp-01-proposal`: `required_conversation_resolution` on, stale-approval dismissal and last-push approval on, force-push and deletion off. `required_approving_review_count` is **0** by your choice, so nothing yet blocks a merge at a reviewed SHA — which is why #3932 could land before these corrections. Flipping that to 1 once a reviewer identity or CODEOWNERS exists is what makes the rule enforceable rather than observed.


mfreeman451 (commented):
**BLOCKING REVIEW** of exact head `41ad46a60156115d6e50c82c7151461b2f8d2c6e`.

GitHub cannot accept `REQUEST_CHANGES` from the authenticated account because it is also the PR author. Treat this commented review as changes requested.

### P1 — the redesign still retains unbounded state on a healthy live lane

`advance_resolved/1` moves every resolved sequence into `disposition` (`resolved_prefix.ex:250-263`), and `release_below/2` is the only deletion path. That function is documented as consuming `EdgeRecordLaneOpen.first_unresolved_sequence`, but the protocol carries `lane_open` only as the first client message; every later client message is a delivery frame (`record.proto:719-725`). The normative scenario explicitly permits one long-lived RPC to carry multiple executions.

Observed at this exact head: after 1,000 in-order authoritative outcomes, `{resolved_through, retained_dispositions, pending_out_of_order} == {1000, 1000, 0}`. Cycling a bounded part-2 in-flight window therefore does not bound the cumulative map.

The structural test does not establish provenance: `release_below/2` accepts a naked integer, and `release_below(t, resolved_through(t) + 1)` succeeds without any agent report. It only blacklists two old function names. Also, on an actual open/resume the required operation is `new(first_unresolved_sequence)`; reusing stale state can reject a legitimate report that is ahead of the stale gateway watermark.

Separate ephemeral gateway-history compaction from agent spool reclaim. Either define a gateway-observable compaction point with a hard resend/history bound, or add a session-bound in-stream agent progress signal. Freeing gateway memory must never authorize agent reclaim, but waiting for a report that cannot arrive in-session recreates the original retain-forever bug.

### P1 — the prior retryable-supersession survivor is still unclosed

The derived matrix puts sequence 2 behind missing sequence 1. For retryable -> audit/quarantine/permanent, replacing `pending[2]` and calling `advance_resolved/1` produce exactly the same asserted state because the leading gap prevents advancement. The only leading-gap supersession test uses authoritative.

A selective mutant can therefore update the pending kind correctly but skip advancement for audit, quarantine, and permanent, wedging the lane while every current assertion stays green. Exercise each resolving kind at sequence 1 with a queued run behind it, and require the exact prefix and disposition map after supersession.

### P1 — release can erase out-of-order PubAck evidence without detection

Every `release_below/2` fixture has an empty `pending` map. A mutation adding `pending: %{}` to the returned tracker satisfies all current release assertions, silently forgetting observed outcomes. Cover release with both retryable and resolving entries above the boundary and assert the complete returned tracker; then fill the gap and prove the retained run collapses.

### P2 — the uint64 boundary is incomplete

`new/1` and `record/3` enforce uint64, but `release_below/2` does not. Observed at this head: resolve `UINT64_MAX`, then release below `UINT64_MAX + 1`; it returns `{:ok, ...}` with `base = 18_446_744_073_709_551_616`, a value the claimed protobuf source cannot express. Reject it or model lane exhaustion explicitly.

### P2 — claims and evidence still need reconciliation

- `invalid_kinds/0` still omits numeric `-1`, `0`, and first-undeclared `6`, although generated enum decoding deliberately preserves negative integers. It also covers only fresh/pending state, not resolved/released state.
- The test moduledoc still says a local watermark in this gateway tracker authorizes deletion, and headings/comments at lines 225-231, 302, and 327 still claim rejected calls prove whole-state equality even though rejected immutable calls return no replacement state.
- “Reclaim now lives with the Go agent” is an ownership decision, not landed runtime code in this two-file PR. The Go package still has zero consumers and still combines both watermarks. Task 3.10 and its required Go/Elixir vectors remain open. Scope the text as scaffolding, and do not also say the only Go implementation should be removed after the Elixir port integrates.

Verification: exact diff is clean; strict OpenSpec validation passes; the uncached Bazel edge shard passed in 54.4s. GitHub checks were still running when this review was submitted.

mfreeman451 (commented):
lgtm

PR #3945, opened 2026-08-24
ci(proto-abi): run the edge suites as Bazel targets instead of 34 sequential steps

mfreeman451:
proto-abi took **20m44s** because it ran 33 `mix test --no-start` steps and one `go test` one after another — no caching, no parallelism.

## Why it was written that way

Not an oversight. **BazelCI's `pull_request` trigger was filtered to `staging`**, so nothing Bazel ran on PRs into `usp-01-proposal`. The suites had to be reimplemented somewhere, and this workflow was the only thing that ran on the stack.

That's the *same* staging-only filter that hid Elixir Quality until it briefly ran by accident during a GitHub stack grouping — and caught eight credo findings the moment it did. Two gates were scoped to a branch this work doesn't target yet.

## What changed

`//buildbuddy.yaml` now triggers on `[staging, usp-01-proposal]`.

Every file the removed steps named is **already a Bazel target** — verified by querying each file's owning target, not assumed:

| what | target |
|---|---|
| 31 edge test files | `//elixir/serviceradar_core:unit_tests_serviceradar_edge` |
| `edge_v1_golden_test` | `//elixir/serviceradar_core:unit_tests_serviceradar_other` |
| Go golden parity | `//proto/edge/v1:edgev1_golden_test` |
| Go edge suites | `//go/pkg/edge/{edgerecord,execstate,projection}:*_test` |

**Measured: 2m22s cold for all six in parallel, against 20m44s sequential.**

They also cover *more* than the list did: a target picks up a new test file automatically, a hand-maintained list of step names does not. A file added to `test/serviceradar/edge/` was previously ungated until someone remembered to add a step — which is how `stream_route_test.exs` and the publisher suite were ungated when first written.

## What stays, and why

Buf lint and the two codegen drift guards need the pinned protoc/Rust toolchain this job installs. The enum-transform fail-closed fixture is a plain script with **no Bazel target** — queried, not assumed — so deleting its step would silently drop the check. Giving it a target is worthwhile follow-up.

**46 steps → 11.**

## Verification

All six targets pass uncached. YAML validated for both files.

## Merge order

Per the current rule I'm not merging integration work ahead of #3938 (the `ResolvedPrefix` correction). This is CI infrastructure rather than integration work, and landing it first would speed up every subsequent round — but that's your call, not an assumption I'll make.

One consequence worth stating: once this merges, PRs into `usp-01-proposal` will start running the **full** BazelCI sweep, which is a much broader gate than proto-abi was. That's the point, and it may surface pre-existing failures on this branch that nothing was checking before.


mfreeman451 (commented):
Blocking review of exact head 6be45061afe3b44efcddec454699f1de77e1726e.

The architecture and merge direction are right: enable BazelCI for the integration branch and delete the duplicate sequential Elixir execution. I found one actual coverage regression and three evidence/comment corrections before this should merge.

P1 — keep Go test discovery, or replace it with an ownership guard. The deleted `go test ./proto/edge/v1/... ./go/pkg/edge/{edgerecord,execstate,projection}/...` discovers future `*_test.go` files and subpackages automatically. The replacement Bazel `go_test` rules enumerate `srcs` by hand (and `proto/edge/v1` is Gazelle-ignored), so a new Go test can now land without entering any target while CI stays green. The claim in `proto-abi.yml` that “a target picks up a new test file automatically” is true for the Elixir globbed shards, not these Go targets. Keep the inexpensive direct Go step for now, or add a source-inventory/drift target in this PR.

P2 — bind the complete moved-test inventory and timing claim. The removed gateway publisher test is owned by `//elixir/serviceradar_agent_gateway:unit_tests_serviceradar_agent_gateway`, but that seventh target is absent from both the comment’s inventory and the reported six-target 2m22 measurement. The live proto-abi run also remained in the Elixir regeneration guard well past 2m22, while BazelCI runs the full repository build/unit/integration/race/benchmark graph separately. Please describe the measured result narrowly (the duplicated test block or selected targets), include the gateway target, and report end-to-end/critical-path timing only after the exact run completes.

P2 — reconcile comments with the new state. The workflow header still says no GitHub workflow covers the whole graph and nothing runs the edge Bazel build; this PR makes BuildBuddy do exactly that for `staging` and `usp-01-proposal`. The two path comments still refer to “the Mix step below,” which is deleted. `buildbuddy.yaml` says roughly 25 Mix steps although the exact removal is 33.

Merge order: after these corrections and an exact-head green BazelCI run, merge #3945 first. Then reconcile #3938 onto the updated integration branch and require a fresh BazelCI result; do not treat its earlier checks as retroactive coverage. Any newly exposed pre-existing failure is useful evidence to triage, not a reason to suppress or bypass the sweep.

Separately, `usp-01-proposal` currently has no required status checks. Once the first baseline is green, making BazelCI required is what turns this from observed coverage into a merge gate.

mfreeman451:
State correction after the review was posted: #3938 is now MERGED at 41ad46a60156115d6e50c82c7151461b2f8d2c6e. The prospective merge-order sentence is therefore no longer actionable. Reconcile #3945 onto the current usp-01-proposal head, close the findings above, and require fresh exact-head Proto ABI and BazelCI results before merging it.

PR #3949, opened 2026-08-24
docs(ci): record the measured proto-abi timing, not the estimate

mfreeman451:
Comment-only. Corrects a claim I left in `proto-abi.yml` that I can now measure precisely.

## The numbers I quoted were both misleading

| claim | reality |
|---|---|
| "20m44s → 2m22s" | 20m44s was an **older** run |
| "12m26s → 2m22s" | 12m26s was right for the job; **2m22s is the six Bazel targets in isolation**, not this job |

**Measured end to end: 12m26s before, 9m28s after — about 24%.**

The mix steps ran against an already-compiled project and were cheap. Toolchain setup, the protoc build, `deps.get` and the two drift guards dominate what remains, which is why removing 34 steps moved the job by three minutes rather than transforming it.

## Why it still mattered

Speed was the smaller win, and I led with it twice. The larger ones:

- These suites now run under **Bazel at all** on this branch — cached, parallel, on the workflow fleet. Before the trigger fix, nothing Bazel ran on PRs into `usp-01-proposal`.
- A **target discovers** a new test file; a hand-maintained list of step names does not. That gap is now also guarded on the Go side by `//build:go_test_inventory_test` (#3948).

The comment records both the measured figures and that ordering, so the next person reading it isn't sold the headline I was.

## Status

BazelCI ran green on #3948 — the first exercise of the new trigger. That was the precondition for making it a **required** check, which is the remaining step to turn it from running into blocking.


mfreeman451:
Post-merge correction required; verified against the GitHub step timestamps rather than the prose.

The central before/after claim is still invalid. The 12m26 run on #3945 and the 9m28 run on #3948 were both POST-migration and had the same Proto ABI step topology. #3948 added the Go inventory guard/comments; it did not remove the sequential Mix block. The exact #3949 run was 9m44, confirming that the roughly three-minute spread is run variance dominated by the Elixir regeneration guard (10m27, 7m36, and 8m18), not a second 24% migration improvement.

The closest exact pre-migration observation is #3938: 36m29 total. Its subsequently removed test block ran from 23:14:22 through 23:36:30, about 22m08, so the statement that those steps were cheap is also contradicted by the run.

Defensible evidence:
- pre-migration observed once: 36m29;
- post-migration observed range: 9m28–12m26 (with #3949 at 9m44);
- the deleted sequential block itself occupied about 22m08 in the pre-migration run;
- BazelCI separately exercised the whole graph successfully.

Please fix forward without assigning a precise causal percentage from one pre sample. Also narrow the comments saying targets automatically discover new tests: that is true for the globbed Elixir shards. Explicit Go go_test srcs do not auto-include a new file; #3948's inventory guard detects the omission and fails, which is different and is correctly described later in the same comment.

PR #3977, opened 2026-08-24
feat(edge): bounded pipelined publish window (task 3.3(c) part 2)

mfreeman451:
Task 3.3(c) part 2 of three. **Pure data and functions — no OTP process**, which stays deferred to part 3 where independently bounded pools create the runtime need.

## What it does

Publishing must be **pipelined** — multiple outstanding un-acknowledged sequences — rather than serializing every frame on one request and waiting for its PubAck. This is the accounting that makes that safe: it bounds outstanding work by frame count, by bytes, and by a per-frame PubAck deadline.

No I/O, no clock. `now` and deadlines are passed in, so the window is deterministic and no test sleeps.

## The bounds come from the GRANT, not a local constant

`granted_frame_credits` (uint32) and `granted_byte_credits` (uint64) arrive in `EdgeRecordLaneOpenAck`, so the window is constructed **from the ack** rather than from configuration that could disagree with the wire. Values outside those ranges are refused — they could not have come from a lane-open ack.

## A deadline REPORTS; only settling releases

The spec says a timeout **republishes the same bytes on the same slot**, so an expired frame is still in flight. If expiry released credits, the window would hand the same budget out twice and real in-flight would exceed the grant — a bound that relaxes exactly when the broker is already struggling.

Investigating your missing-mutation point showed this invariant is **structural, not behavioural**: `expired/2` returns `[pos_integer()]`, so it has no channel through which to return released credits. What was actually missing was a pin on *that* — the return shape — rather than another assertion.

## The public surface is an exact inventory

Your fail-open point was correct. Matching names on `release`/`drop`/`reclaim` would wave through `settle_timed_out/2`, which releases credits and contains none of those words. The test now compares the **exact exported function set**, so any new public function fails until someone classifies whether it releases. `settle/2` is the only one permitted to.

## Transition policy, stated once

```
admit a new sequence, within both bounds   -> {:ok, window}
admit a sequence already outstanding       -> {:error, :already_outstanding}
admit beyond the frame grant               -> {:error, :frame_credits_exhausted}
admit beyond the byte grant                -> {:error, :byte_credits_exhausted}
settle an outstanding sequence             -> {:ok, window}
settle anything else                       -> {:error, :not_outstanding}
```

`settle/2` deliberately does **not** distinguish "never admitted" from "already settled" — telling them apart would require retaining every settled sequence forever, which is the unbounded growth this bounds. The docstring says so rather than leaving a caller to infer it.

## Verification

**Seven behaviour-changing mutants killed:**

| mutation | |
|---|---|
| byte grant unenforced | KILLED |
| frame grant off-by-one | KILLED |
| re-admitting an outstanding slot | KILLED |
| settling an unknown sequence succeeds | KILLED |
| `settle/2` releases all bytes, not that frame's | KILLED |
| deadline `<` instead of `<=` | KILLED |
| a releasing function added to the public surface (`settle_timed_out/2`) | KILLED |

Removing a public function is caught too.

**One behaviourally equivalent control remained green** — a release computed then discarded, which immutability makes inert — and is **excluded from the mutation score** rather than reported as a kill.

Also pinned per your list: settling one of differently-sized frames releases exactly that frame's bytes; a second settlement returns `:not_outstanding` without releasing again (asserted via observable capacity, not by comparing an immutable input to itself); a fully-expired full window still admits nothing, and settling one frame admits exactly one replacement; and the documented `deadline <= now` boundary.

774 tests, 0 failures. Credo-clean, format-clean.

## Next

Part 3 — separately bounded bulk/interactive/recovery pools, where the OTP process becomes justified and "state unchanged after a rejected call" becomes observable and worth testing.


mfreeman451 (commented):
Review of exact head `8ea2a8d86ae7881119a56a0f05c6c03faabd3374` against current `usp-01-proposal`.

The pure data/function boundary is the right one for part 2; there is still no runtime reason for an OTP process until part 3. The accounting operations themselves are small and readable. I ran the exact Elixir edge Bazel shard uncached, `mix format --check-formatted`, and `git diff --check`; all passed. BazelCI and proto-abi are green at this SHA.

Blocking findings:

1. **[P1] `new/2` freezes lane-handshake behavior that this task does not own, and it conflicts with the recorded contract.** The moduledoc and tests say the two raw integers came from `EdgeRecordLaneOpenAck`, accept zero as a real grant, and accept the full uint32/uint64 ranges. But freeze task 1.7-e is still open and explicitly owns both halves of that negotiation: zero grants refuse, grants satisfy `1 <= granted <= requested`, and the current Go hard caps are `1 << 20` frames and `1 << 30` bytes. Since `new/2` receives neither the request nor the ack, it cannot establish provenance or the `granted <= requested` relation. Please keep this PR a generic accounting primitive over **already validated** negotiated credits and remove the ACK/wire-validity and zero/max claims, leaving the request/ack binding to 1.7-e and the part-3 owner. Alternatively, closing 1.7-e requires its normative/shared peer evidence; it should not be silently decided here.

2. **[P1] An expired retry has no way to acquire a new PubAck deadline.** `expired/2` reports a sequence forever, while `admit/4` rejects the same sequence as `:already_outstanding`. Settling and re-admitting would briefly release the very budget this design says a retry must retain. Reproduction: admit sequence 1 at deadline 100; it is expired at 100; `admit(w, 1, ..., 200)` refuses; it is still expired at 101. Add a pure retry/rearm transition that changes only the deadline of an existing outstanding sequence, preserves its exact frame/byte charge, and refuses an unknown sequence. No process is needed for this.

3. **[P1] The “exact exported surface” guard is fail-open for exported macros.** `PublishWindow.__info__(:functions)` does not include `defmacro`/`defguard`; an added `settle_timed_out/2` macro can expand to a releasing path while the current guard stays green. Bind `__info__(:macros)` exactly as well (currently it should be empty), so every public callable form is classified.

Additional corrections:

- **[P2] `settle/2`'s contract is too broad.** “A terminal disposition that ends the attempt” can be read as allowing classification alone to release credits. Task 3.5 says primary, audit, quarantine, security-quarantine, and permanent-rejection outcomes resolve only after the required destination PubAck. State that authoritative PubAck precondition here.
- **[P2] The expiry invariant is module-local, not structurally closed end to end.** A list return proves only that `expired/2` cannot directly return a modified window; the future owner can still pass every expired sequence to existing `settle/2`. Part 3 needs a mutation/transition test on the timeout branch proving it republishes/rearms while the window remains charged. Narrow the stronger comments in this PR accordingly.
- **[P2] `admits?/2` disagrees with `admit/4` for malformed input.** `admits?(w, -1)` returns true in a fresh window while `admit(w, ..., -1, ...)` returns `{:error, :bytes}`. Make the predicate false for non-integers and negatives.
- **[P2] The interleaved conservation test checks only the final state.** The `tap/2` runs after the complete reduce, so a transient overcommit that later settles away passes. Assert frame/byte bounds after every transition and explicitly assert the intended over-capacity admission was refused.
- **[P2] “Charged by actual size” is not established here.** `admit/4` accepts a caller-supplied scalar; the present tests stay green if a future owner charges `1` for a large publication. Either derive the charge from the publication at this boundary or state that this is a generic weighted counter and make exact encoded-byte binding a required part-3/3.4 integration test.
- **[P2] Deadline policy is likewise supplied, not enforced.** This module pins `<=` and stores an arbitrary monotonic instant, but it does not establish the configured PubAck duration/unit promised by the proposal. Say that explicitly and bind it where the publisher constructs deadlines.
- **[P3] Reconcile the mutation count.** The evidence says seven behavior-changing mutants were killed and then separately says deleting a public function was caught. If that deletion was actually applied, count it as an eighth behavior-changing/structural experiment or label it explicitly as an unscored control.

The core accounting can remain pure. The blockers are contract ownership, the missing retry transition, and the incomplete public-surface inventory—not a request to add OTP early.

PR #3985, opened 2026-08-24
fix(edge): stop freezing 1.7-e's handshake semantics; make the retry representable

mfreeman451:
Corrections to `8ea2a8d86a`, which merged as #3977 before these were ready. Reconciled onto current `usp-01-proposal` (0 behind).

## P1 — `new/2` froze semantics task 1.7-e owns

You're right and I had this backwards. 1.7-e owns the lane-handshake admission bounds, they are **not yet frozen**, and its return relation is **`1 <= granted <= requested`** — so a zero grant is a *refusal*, not "a real answer" as I documented. The uint32/uint64 caps I enforced were my invention.

1.7-e even names this failure mode directly: a different semantics *"SHALL be chosen and stated HERE, not discovered from whichever verdict the current Go code returns."* Inventing caps inside an accounting module is exactly that.

`new/2` now checks **type preconditions only**. `ValidateLaneOpenAck` adjudicates legality. A zero grant is accepted and simply has no capacity — stated explicitly as *not* a claim that zero is legal.

## P1 — the retry path was unrepresentable

An expired frame must be republished with a **new** PubAck deadline. But `admit/4` refuses an outstanding slot (`:already_outstanding`) and `settle/2` would release credits for a frame that is still in flight — so an expired frame could be reported forever and never re-armed.

`rearm/3` moves the deadline only: charges nothing, releases nothing, refuses a sequence that is not outstanding. There's a test that re-arming creates **no** capacity, since that is how it could quietly become a release.

## P1 — the public-surface guard ignored macros

`__info__(:functions)` omits macros entirely, so a `defmacro release_expired/2` would have added a releasing entry point the inventory could never see. It now compares **functions and macros**, with `@public_macros []` asserted empty rather than assumed.

## Smaller contract/test gaps

- **`settle/2` precondition**, now stated: a *validated* PubAck, or a terminal disposition. An ack from an unexpected stream, or one that failed sequence validation, is not a settlement — releasing on it returns budget for a frame whose fate is unknown.
- **`admits?/2`** answers `false` for a malformed size rather than raising.
- **Transient overcommit** is checked after *every* operation, not only at the end — a window that overcommitted then corrected itself would pass an end-state assertion.

**Still owed, and not claimed:** binding `bytes` to the actual encoded frame size, and the deadline policy itself (who chooses the interval). Both belong with the publisher that drives this.

## Verification

**Four mutants killed on the new surface**, plus the seven from the previous head:

| mutation | |
|---|---|
| a releasing `defmacro` (invisible to the old guard) | KILLED — failure names it |
| `rearm/3` releases credits instead of moving the deadline | KILLED |
| `rearm/3` admits an unknown sequence, bypassing both bounds | KILLED |
| `admits?/2` raises instead of answering false | KILLED |

One process note worth recording: the `rearm`-releases mutant first reported SURVIVED because the formatter had rewritten the target line, so the substitution never applied. **A mutation that does not apply is a failed experiment, not a pass** — re-run against the real text, it killed. That's the third time formatter rewriting has broken a position-dependent edit for me; reading the file back first is the fix.

780 tests, 0 failures. Credo-clean, format-clean.


mfreeman451 (commented):
Review of exact head `d8beb216e7f4d0462d175e27884c2119a9e354b7` against GitHub base `e15ac7afa3267fc8ffe64e447ca21f69dd421d6a`.

The pure accounting boundary remains the right design. `rearm/3` now makes retry representable without releasing frame or byte capacity, `admits?/2` handles malformed input, and the focused Bazel shard passes at 780 tests. The following blockers remain.

1. **[P1] B3 is still fail-open because function and macro kinds are erased before comparison.** The test concatenates `__info__(:functions)` and `__info__(:macros)`, then compares the combined `{name, arity}` tuples. Replacing a classified function with a same-name/same-arity macro therefore leaves `actual` unchanged. I reproduced it independently: a module with `def rearm/3` and one with `defmacro rearm/3` both produce the same combined inventory, `[rearm: 3]`. The added-macro mutant proves a new name is caught; it does not prove kind substitution is caught. Compare the function list and macro list independently, or tag every entry with its kind.

2. **[P1] B1 still describes an upstream validation boundary that does not exist, and the executable evidence retains the old zero-credit claim.** The module says capacities come from an “ALREADY-VALIDATED” `EdgeRecordLaneOpenAck` and that `ValidateLaneOpenAck`/“the validator upstream” rejects illegal grants. There is no Elixir `ValidateLaneOpenAck`; task 1.7-e explicitly says the Elixir peer is still owed and no ingress reaches even Go's validator. `new/2` receives two bare integers. Also, `publish_window_test.exs:20-23` still says credits come from the lane-open Ack and zero is “a real answer,” while lines 30-40 correctly say zero is a refusal. Accepting arbitrary nonnegative capacities is fine for a generic bignum accounting helper; describe them as caller-supplied capacities and say the future 1.7-e/part-3 boundary must supply a validated request/Ack pair. Remove the present-tense provenance and stale zero claim.

3. **[P1] The test suite now contains the exact timeout-means-release wiring the design prohibits.** The new loop at `publish_window_test.exs:382-386` takes `[oldest | _]` from `expired/2` and immediately calls `settle/2`, with no PubAck evidence. That directly contradicts “expiry reports; only a validated publication outcome releases.” It also disproves the surviving claim at lines 125-132 that the invariant is structurally closed: the list return cannot mutate the window, but a caller can feed the list into existing `settle/2`. Cycle the test with explicit simulated validated-PubAck operations, never by deriving settlement eligibility from expiry. Narrow the structural claim to `expired/2`'s local return shape, and leave the owner timeout/retry mutation for part 3.

4. **[P1] `settle/2` still permits a bare terminal disposition.** Its doc says the caller may settle after a validated PubAck “or a terminal disposition that ends the attempt.” Task 3.5 requires the primary/audit/quarantine/security-quarantine/reject-audit destination PubAck before each terminal outcome resolves; classification alone is insufficient. If a failure ends before publication, it should never have entered this publish window. Require the validated PubAck for the selected publication outcome and explicitly exclude bare internal/wire disposition classification.

Smaller reconciliations:

- **[P2]** Add `rearm/3` to the “Transition policy, stated once” table. It currently accepts a non-expired sequence and an equal/backward deadline; that can remain owner-level policy, but the API must state that eligibility and future/monotonic deadline validation are deferred to the publisher.
- **[P2]** The test still says bytes are “charged by actual size,” although it supplies an unrelated integer and the PR says actual encoded-size binding is owed. Rename this to the caller-supplied charge and point at the part-3/3.4 binding obligation.
- **[P2]** “A slot is allocated once” overstates what this forgetful window enforces: `admit -> settle -> admit` for the same sequence succeeds. Say “not twice concurrently”; historical sequence reuse belongs to the owner/ResolvedPrefix.
- **[P2]** The new “after every operation” test asserts after admission but performs settlement afterward without another assertion; the older end-only interleaving test also remains. Use one transition helper that asserts the complete bounds after every returned state and prove at least one admission and one refusal occurred.
- **[P3]** Mutation accounting still says four new plus the prior seven while the prior head separately claimed a public-function deletion probe. Classify that deletion as a scored behavior-changing/structural mutation or explicitly as an unscored guard control.

Verification at this exact SHA: `mix format --check-formatted`, both relevant OpenSpec strict validations, `git diff --check`, Bazel ownership query, and uncached `//elixir/serviceradar_core:unit_tests_serviceradar_edge` all pass. No OTP/process work is requested here.

PR #4003, opened 2026-08-24
fix(edge): settle by allowlist, per-kind surface guard, no settlement from expiry

mfreeman451:
Corrections to `d8beb216e7`, which merged as #3985 before these were ready. Reconciled onto current `usp-01-proposal` (0 behind).

## P1 — function↔macro kind substitution passed

The guard compared `functions ++ macros` as **one merged set**, so `def settle/4` → `defmacro settle/4` left `{:settle, 4}` present either way while changing call semantics to compile-time expansion. Compared **per kind** now, with separate inventories and separate failure messages.

Verified: the substitution is **KILLED**, reporting `exported FUNCTIONS drifted: added [do_settle: 4], removed [settle: 4]`.

## P1 — the Elixir ACK validator does not exist

The moduledoc said `ValidateLaneOpenAck` decides grant legality. **That's Go.** This runtime has only `SemanticValidate.validate_lane_open/1` — an enum/shape check on the *request* half — and no ack validator at all. So **nothing on this side validates a grant before it reaches this window.**

The moduledoc now states that, names 1.7-e as owing the peer, and explains why this module still refuses to substitute for it. Same class of unbacked claim as the earlier "reclaim lives with the Go agent".

## P1 — a test settled directly from `expired/2`

The overcommit test picked the oldest **expired** frame and settled it — modelling exactly the caller behaviour the module forbids, while looking like coverage. It now settles on a simulated ack and never on expiry.

## P1 — `settle/4` allowed disposition without its PubAck, and was fail-open on kinds

Two defects, one of which you predicted and I hadn't fixed:

- **PubAck** was prose in a docstring, so nothing enforced it. It's an argument now: a resolving disposition without one is `:pub_ack_required`, and `REJECTED_RETRYABLE` doesn't settle at all — transient, stays charged, re-arm instead.
- **Kind check was fail-open.** It asked the generated mapping *"is this declared?"*. A future enum member is neither unspecified nor retryable, so it would fall through to a release and hand back credits for an outcome nobody classified. It's an **allowlist** of the four known resolving kinds now.

## Mutation accounting

**Three killed** on this surface: kind substitution · PubAck requirement dropped · retryable allowed to settle.

**One survived, and it is an equivalent mutant** — reported rather than dressed up. Reverting the allowlist to the mapping check is behaviourally identical with today's five-member enum; the two differ only when the proto grows. The protection for that case is the derived test asserting the settling set is **exactly four** *and* that **five kinds are declared**, so the suite fails the moment a member is added and forces classification.

## Against your verification list

| check | result |
|---|---|
| function↔macro substitution fails by kind | ✅ killed, drift named |
| no present-tense Elixir ACK-validation claim | ✅ only remaining mention is the *denial* |
| no test or API path derives settlement from expiry | ✅ none |
| `settle/4` requires PubAck for every resolving kind | ✅ all four exercised |
| retryable stays charged | ✅ |
| zero/future enum members fail closed | ✅ allowlist, plus the derived set test |

785 tests, 0 failures. Credo-clean, format-clean.


mfreeman451 (commented):
Reviewed exact head `67970be2fbf8650764dd2c55d6234ad2e4019f73` against base `e12746e0f87c7d18c48adbd942adfba41db1b222`.

The per-kind function/macro inventory, retryable charge retention, four-kind fail-closed allowlist, and removal of the test's expiry-derived settlement all check out. Two P1 blockers remain.

### P1 — `settle/4` still accepts a sentinel, not the required PubAck

`publish_window.ex:187-206` types `pub_ack` as `term()` and rejects only `nil`. I reproduced `false`, `%{}`, `:timeout`, `{:error, :timeout}`, and `%{stream: "WRONG", seq: 0}` all returning `{:ok, window}` and releasing the stored charge. A copied success token likewise settles a different edge sequence.

The test normalizes that hole at `publish_window_test.exs:18-20,442-452`: one atom-keyed bulk-stream stand-in with `seq: 1` settles authoritative, audit, quarantine, and permanent outcomes. Task 3.5 instead requires the PubAck from the publication's actual primary/audit/quarantine/security-quarantine/reject-audit destination. Those are six internal outcomes mapped onto five wire dispositions; in particular, the two quarantine routes collapse onto the same disposition, so `disposition` cannot derive which destination had to Ack.

This can return credits while the real publication is still unresolved, defeating the hard window. The accounting boundary needs attempt-correlated validated settlement evidence carrying the internal outcome/expected destination (minted and consumed by the publisher owner), or release must remain private/deferred until that owner exists. At minimum, error, malformed, wrong-attempt, and wrong-destination evidence must not release. A non-`nil` `term()` is no stronger than the prose precondition it replaced.

### P1 — the nonexistent Elixir lane-Ack validator is still asserted in the present tense

The new caveat at `publish_window.ex:32-37` correctly says this runtime has no lane-open-Ack validator, but `new/2` immediately calls its input an “ALREADY-VALIDATED `EdgeRecordLaneOpenAck`” and says “the validator upstream” refuses zero at `publish_window.ex:90-100`. The test repeats “ValidateLaneOpenAck adjudicates” at `publish_window_test.exs:45-55`, and `publish_window.ex:28` still assigns legality to “the lane-open validator.” Task 1.7-e explicitly records that the Elixir peer does not exist and that even Go's validator has no live ingress caller.

Until 1.7-e lands, describe these as caller-supplied accounting capacities and remove the Ack provenance/upstream-enforcement claims. The PR body's statement that only the denial remains is false at this head.

### P2 — reconcile the executable contract and its evidence

- The module still documents nonexistent `settle/2` at `publish_window.ex:47,59,229`, and the transition table omits the new retryable/unknown/missing-PubAck/rearm outcomes. The inventory comment repeats the old arity at `publish_window_test.exs:185-188`.
- `publish_window_test.exs:37` still calls zero “a real answer” while the next test correctly says it is a 1.7-e refusal.
- The “after EVERY operation” test at `publish_window_test.exs:379-410` asserts bounds after admit but not after the optional settlement. Assert on every returned transition state, including settlement/rearm.
- `publish_window_test.exs:202-208` is vacuous as an unchanged-state test: immutable `w` cannot be mutated by the call, and `expired/2` returns no window. The capacity assertions and return-shape test are the load-bearing evidence; rename/delete this one rather than count it as state evidence.
- The test calls a caller-supplied integer “actual size” at `publish_window_test.exs:88`, and calls `deadline <= now` “the contract” at `:246-253`, although encoded-size binding and deadline policy are explicitly deferred to the publisher. Keep those claims scoped to caller-supplied charge/deadline until the owner binds them.

Independent verification: focused uncached Bazel shard passed; `mix format --check-formatted`, `openspec validate unify-sweep-results-proto --strict`, and `git diff --check` passed. The green suite does not exercise the reproduced non-`nil` sentinel path.

PR #4008, opened 2026-08-24
fix(edge): require real PubAck evidence and the six-valued internal outcome

mfreeman451:
Corrections to `67970be2`, which merged as #4003 before these were ready. Reconciled onto current `usp-01-proposal` (0 behind).

## P1 — `settle/4` accepted any non-nil term as PubAck evidence

You reproduced it exactly. `not is_nil(pub_ack)` was the whole check, so `false`, `{:error, :timeout}`, bare atoms and malformed maps all **released credits for frames whose fate was unknown**.

It now requires the shape `JetStreamPublisher.parse_ack/1` returns for a success — a non-empty binary `stream` and a positive integer `seq`. Fourteen bogus forms are tested, including the ones you named.

The limit is stated rather than implied: this module still cannot judge **authenticity**, or whether the ack came from the destination this outcome required — the publisher fences that against the resolved route. What it can refuse is evidence that is not a PubAck at all.

## P1 — it took the wire disposition, which cannot identify the required destination

Task 3.5 has **six internal outcomes and five wire members**: `quarantine_publication` and `security_quarantine_publication` both map to `ACCEPTED_QUARANTINE`. Taking the wire member made the destination unidentifiable — a security-quarantine record could settle on a plain quarantine PubAck.

`settle/4` now takes the internal outcome. `wire_disposition/1` exposes the five-to-six mapping, with a test asserting both quarantines report the same wire member while remaining **distinct internally** — which is what keeps the security-quarantine routing path recoverable.

## P1 — present-tense lane-Ack validation claims

*"So legality is the lane-open validator's"* sat two paragraphs above *"THIS RUNTIME HAS NO PEER."* The moduledoc asserted and denied the same thing.

It now opens with **"NOTHING VALIDATES THE GRANT ON THIS SIDE TODAY"** and records that the earlier wording read as though something upstream checked it. My previous sweep grepped for phrasings I expected instead of reading the text, which is why I reported it clean — that was my error, not a subtle case.

## Smaller items

- Three surviving `settle/2` references in the transition policy.
- The zero-grant paragraph read as though zero were legal. Zero is not *rejected* here because this module doesn't adjudicate grants at all — not because zero is admissible.
- The two overstated tests were replaced: the enum-derived "settling set" now enumerates **outcomes** (including a `:some_future_outcome` probe), and the disposition test now asserts a wire member is explicitly *not* an internal outcome.

## Verification — 3 mutants killed

| mutation | |
|---|---|
| any non-nil term accepted as PubAck evidence | KILLED |
| outcome allowlist dropped, any atom settles | KILLED |
| `security_quarantine` collapsed onto the primary disposition | KILLED |

787 tests, 0 failures. Credo-clean, format-clean, `openspec validate --strict` valid, `git diff --check` clean.


mfreeman451 (commented):
Reviewed exact head `f55aa1047d2b13fe056c06d47890a0ffee4df7d6` against base `8c938916d0040a0747c0bd74a519d6f95f3fdfc6`.

The six internal names and their current five wire values are correct; retryable remains charged; unknown outcomes fail closed; and the basic non-map/sentinel cases are now refused. The prior correlation blocker is still open, however, and several claimed reconciliations are not present at this head.

### P1 — the Ack is still detached from the publication it allegedly settles

Task 3.5 requires the Ack for the exact record identity and the outcome-specific primary/audit/quarantine/security-quarantine/reject-audit destination (`specs/ingestion-routing/spec.md:152-169`). `PublishWindow` stores only `{bytes, deadline}` for an edge sequence (`publish_window.ex:90-95`); `settle/4` takes an independently supplied outcome and raw map, while `usable_pub_ack?/1` checks only a nonempty stream and positive integer (`:202-230`). It never compares the outcome to the Ack stream, and it has no request/attempt/record identity to compare.

I reproduced one bulk-stream Ack settling `:security_quarantine_publication`, then reusing that same Ack to settle a second lane sequence as `:permanent_rejection`; both returned `{:ok, ...}` and the window drained completely. The committed test codifies the same bypass at `publish_window_test.exs:443-453`: one bulk Ack settles all five terminal outcomes.

`JetStreamPublisher` fencing does not survive detaching the returned map from its request context. The Ack's `seq` is the JetStream stream sequence, not the edge lane sequence, so those cannot simply be compared either. There are no non-test `PublishWindow` callers yet, and the publisher explicitly says its DLQ publications do not exist, so there is currently no owner-level integration evidence closing the handoff.

This still needs evidence minted from the exact successful async publication context and bound to the edge sequence, internal outcome, expected destination, and record identity—or equivalent expectations stored in each outstanding attempt and verified during settlement. If that properly belongs to the part-3 publisher, keep this helper explicitly accounting-only and leave the task-3.5 enforcement claim open until integration; an Ack-shaped map is not the required PubAck proof.

The new shape predicate is also weaker than the parser it cites. `JetStreamPublisher.parse_ack/1` caps `seq` at u64 and normalizes a boolean `duplicate` (`jet_stream_publisher.ex:53,219-250`); `usable_pub_ack?/1` accepts an over-u64 sequence, missing/invalid `duplicate`, and even an error-bearing map. I reproduced `%{stream: "S", seq: 2^64, duplicate: "invalid", error: :timeout}` releasing credits. Tighten the exact parser-output shape if raw maps remain, but that alone does not solve correlation.

### P2 — the six-to-five proof is not exact or bidirectional

`@settling_outcomes` currently has the right literals, but the tests do not pin the complete normative mapping. `publish_window_test.exs:485-501` checks only both quarantines plus retryable; primary, audit, and permanent can be remapped with that test still green. The “EXACTLY five” test at `:457-483` walks a hand-written candidate list, so adding a differently named production outcome does not enter the test despite its comment claiming it will.

Bind one exact six-row `{internal outcome, generated wire member}` set bidirectionally to a discoverable production inventory and to `Serviceradar.Edge.V1.EdgeRecordDispositionKind.mapping()`. Today a generated enum rename/removal also leaves this suite green because it compares hard-coded atoms only.

### P2 — the prior documentation/test findings remain at this head

- The moduledoc correctly says no Elixir lane-Ack validator exists at `publish_window.ex:28-37`, but `new/2` still says “ALREADY-VALIDATED `EdgeRecordLaneOpenAck`” at `:98-109`; the test still says `ValidateLaneOpenAck adjudicates` at `publish_window_test.exs:45-50`.
- `publish_window_test.exs:37` still calls zero “a real answer,” contradicting the refusal wording immediately below it.
- Exact `rg -F 'settle/2'` across the module and test returns two stale test references, at lines 187 and 322. The final commit fixed the library docstring, not the last repository occurrence.
- The two tests identified in the prior review were not replaced: the “after EVERY operation” test still performs settlement after its last assertion (`:380-411`), and `expired/2 is a pure report: the window is unchanged by asking` remains the same immutable-input comparison (`:203-209`). The new outcome tests address different properties.
- “charged by actual size” at line 88 and the caller-supplied deadline described as “the contract” at `:247-255` remain overstated while size binding and deadline policy are deferred to the publisher.

Independent verification: the focused uncached Bazel shard passed; both relevant OpenSpec changes validate strictly; `mix format --check-formatted` and `git diff --check` passed. The runtime reproductions above are not covered by the green suite.

PR #4011, opened 2026-08-25
fix(edge): settle is accounting-only; the Ack-shaped map was not proof

mfreeman451:
The correlation blocker against `f55aa1047d` (merged as #4008 before this was ready). Reconciled onto current `usp-01-proposal`, 0 behind.

## P1 — you're right, and I took the option you offered

An Ack detached from its publication context proves nothing. Task 3.5 requires the PubAck for the **exact record identity** and the outcome-specific destination. This module has neither: an outstanding entry is `{bytes, deadline}` against an edge lane sequence — no record identity, no attempt token, no expected destination.

Your reproduction was already **codified in my own test**: one bulk-stream ack settled all five terminal outcomes, and the same ack could settle a second lane sequence under a different outcome. And as you note, the ack's `seq` is the **JetStream stream sequence**, not the edge lane sequence, so they can't be compared even in principle.

**The parameter is removed, not tightened.** `settle/3` is accounting-only and says so: the caller must have verified the PubAck, its destination, and its record identity — none checkable here. **Task 3.5's enforcement is owed by the part-3 publisher and remains open.** A weaker imitation of proof standing in front of a real gap is worse than an acknowledged gap, and tightening the shape predicate would have produced exactly that.

## P2 — the six-to-five proof was neither exact nor bidirectional

It checked both quarantines plus retryable, so primary, audit and permanent could be remapped with it green — and it compared hard-coded atoms, so a rename or removal in the generated enum left it green too.

It now pins **all six rows** and asserts the five distinct wire targets are **exactly** the generated enum's non-unspecified members, so upstream drift fails here rather than leaving literals agreeing with themselves.

## P2 — findings that survived my earlier sweeps

Every one of these was reported fixed by me and wasn't:

- `new/2` still said **"ALREADY-VALIDATED"**
- a **third** present-tense `ValidateLaneOpenAck` claim sat in its docstring
- **two `settle/2` references remained in the test file** — I'd fixed the library and declared it done
- **"a real answer"** still contradicted the refusal wording immediately below it
- the **"after EVERY operation"** test settled *after* its final assertion, so that settlement was never checked — it settles first now
- **"`expired/2` is a pure report"** compared an immutable input to itself and could not fail; it now asserts observable **capacity**, with a non-vacuity check that settling *does* move it
- **"charged by actual size"** and the deadline **"contract"** are marked deferred: nothing binds `bytes` to the encoded frame size, and this module doesn't own the deadline policy

I verified this round by **counting occurrences** (`grep -c` across both files) rather than checking the places I expected drift. That's what found the third validator claim and the two test-file `settle/2` references — my previous sweeps sampled and reported clean.

785 tests, 0 failures. Credo-clean, format-clean, `git diff --check` clean.


mfreeman451 (commented):
Reviewed exact merged head `61af21373cfe91764e4f9aabce3654c09b8a0ade` against PR base `19f5d581c6c2ee58f5cd3431f36a13e0b35d2f8a`.

The central correction is sound. `settle/3` is now honestly accounting-only, has no detached Ack-shaped evidence channel, preserves retryable charges, fails closed on unknown outcomes, and the six-internal-to-five-wire mapping is exact against the generated enum. I found no new production/runtime blocker.

There are still four correction groups at this exact head:

1. **[P2] The replacement expiry no-drain test is still vacuous under immutability.** At `publish_window_test.exs:201-218`, all twenty `expired/2` return values are discarded and capacity is reread from the same immutable `w`. No call can mutate that caller binding, so those assertions cannot fail because of what `expired/2` did. Showing that `settle/3` returns a different term does not make the earlier same-input comparisons non-vacuous. The full-window test at `:120-135` and the structural return-shape test at `:137-151` are the real evidence; remove or accurately rename this one.

2. **[P2] The promised per-transition bound check still occurs before settlement.** `publish_window_test.exs:389-420` says it checks after every operation, but the assertions are at `:405-406` and settlement happens afterward at `:412-416`; the returned settled state is not asserted before the next operation. The other interleaved test checks only after the entire reduce. Compute each operation's returned state first, then assert both bounds immediately on that state.

3. **[P2] Encoded-size binding is still claimed where it is explicitly absent.** The production doc correctly says `bytes` is merely caller-supplied and unbound until publisher integration (`publish_window.ex:134-142`), but the test at `publish_window_test.exs:85-95` is still named “charged by actual size” and only passes literal integers. Rename it to supplied/accounted bytes. Also name formal owner task 3.4 (`tasks.md:748-757`) alongside the remaining 3.3(c) integration instead of only the informal “part-3 publisher,” so the obligation cannot get lost.

4. **[P2/P3] Reconcile the remaining task-3.5/stale wording before the publisher consumes it.** The authoritative task/spec list the SECURITY-quarantine PubAck independently, but `design.md:252-257` and `design.md:1001-1004` still collapse it into/omit it from their exhaustive summaries; task 3.5 explicitly calls that omission a prefix-wedge hazard. In the test, `:185` says `{:settle, 2}`, `:194` still calls `settle/4` the sole releaser, and `:408-414` says it settles “ON AN ACK” although this PR deliberately removed that evidence channel. Finally, the transition table's “settle anything not outstanding -> :not_outstanding” is true only for a known settling outcome because outcome validation runs first; qualify it or state precedence.

Independent verification at this SHA: focused uncached `//elixir/serviceradar_core:unit_tests_serviceradar_edge` passed; `mix format --check-formatted`, `openspec validate unify-sweep-results-proto --strict`, and `git diff --check` passed.

PR #4013, opened 2026-08-25
fix(edge): remove the vacuous expiry test; assert bounds per transition; reconcile wording

mfreeman451:
The four correction groups against `61af2137` (merged as #4011 before these were ready). Reconciled onto current `usp-01-proposal`, 0 behind.

## 1 — the replacement no-drain test was still vacuous

You're right that renaming it didn't fix it. All twenty `expired/2` returns were discarded and capacity was reread from the same immutable binding, so **nothing the function did could make it fail** — and showing that `settle/3` returns a different term doesn't rescue the earlier same-input comparisons.

**Removed rather than renamed.** The full-window test and the structural return-shape test are the real evidence; a third test that cannot fail only inflates apparent coverage.

## 2 — the per-transition bound check still ran before settlement

It asserted after the admit and settled afterwards, so the settled state went unchecked until the next iteration — a settlement that overcommitted was never seen at the point it happened. Both bounds are now asserted on the state **each operation actually returned**.

Worth stating: **my earlier fix for this never applied.** The formatter had rewritten the target text and the substitution silently missed, so the old structure survived into your review. Same failure mode as the `rearm` mutant that falsely reported SURVIVED.

## 3 — encoded-size binding was still claimed in a test name

"charged by actual size" promised a binding this module doesn't have. Renamed to the **supplied** byte count, and the formal owner is now named — **task 3.4** requires publishing the exact `record_bytes` and bounding publication by encoded bytes. `settle/3`'s open obligation likewise names **task 3.5** rather than only "the part-3 publisher", so neither can dissolve into informal prose.

## 4 — security-quarantine omitted from two exhaustive design summaries

`design.md` listed four resolving PubAck kinds in both its watermark summary and its publisher step list, omitting the **security-quarantine DLQ PubAck**. Task 3.5 calls that exact omission a prefix-wedge hazard: both quarantine variants collapse onto one wire member, so leaving it out lets a compromise-revoked record reach its DLQ and then **pin the resolved prefix forever**. Restored to both, with the reason recorded at the second.

Also reconciled: a stale `{:settle, 2}` and `settle/4` in the surface-guard comments, "SETTLING ON AN ACK" describing an evidence channel this series removed, and the transition table's last row — which holds only for a known settling outcome, since outcome validation runs first. **Precedence is now stated.**

## Verification

Checked **by count, not by sampling** — the habit that let items survive three of your reviews:

| pattern | occurrences |
|---|---|
| `settle/4` | 0 |
| `settle/2` | 0 |
| `{:settle, 2}` | 0 |
| `ON AN ACK` | 0 |
| `charged by actual size` | 0 |
| `part-3 publisher` (as sole owner) | 0 |
| security-quarantine in design summaries | 2 |

784 tests, 0 failures (one fewer — the vacuous test is gone). Credo-clean, format-clean, `openspec validate --strict` valid, `git diff --check` clean.

## Still open, unchanged

PubAck correlation (task 3.5 + 3.3(c) publisher integration), encoded-size binding (task 3.4), and deadline policy. None claimed as done.


mfreeman451 (commented):
Reviewed exact head `3f7bf86c3eb0df7062f33c37734ad931fc6b319b` against base `05921e292dac988b3fbcafd1222d7834d7e739f3`.

The vacuous expiry test is gone; the dedicated 40-step invariant now checks the returned post-admit and post-settle states immediately; formal ownership is accurate; and both security-quarantine summaries are reconciled. No production/runtime blocker found.

One prior P2 remains in a second test:

**[P2] The separate interleaved-run test still checks only the final state.** `publish_window_test.exs:542-579` is titled “outstanding bytes never exceed the grant across an interleaved run,” but the reducer at `:558-570` performs every admit and settle with no assertion. Bounds are checked once after the complete reduction at `:571-574`, and `:578-579` then requires that final state to equal the empty initial window. Consequently those bound assertions are implied by the final equality and cannot detect a transient overcommit that a later transition corrects. This is the “other interleaved test” called out in the previous review. Assert both bounds on each operation's returned state inside the reducer, or rename/remove this test now that the corrected 40-step test carries the per-transition property.

Independent verification: `mix test --no-start test/serviceradar/edge/publish_window_test.exs` passed 31/31; formatting, strict OpenSpec validation, and `git diff --check` passed. All exact-head PR checks, including BazelCI, are green.

PR #4018, opened 2026-08-25
test(edge): name the interleaved test for the property it actually proves

mfreeman451:
The last P2 against `3f7bf86c` (merged as #4013 before this was ready). Reconciled onto current `usp-01-proposal`, 0 behind.

## The bound assertions were implied by the final equality

You're right. The test asserted bounds only on the **final** state — which the next line shows is an **empty** window, so `0 <= 700` and `0 <= 2` are trivially true. They could never detect a transient overcommit that a later transition corrected, and the 40-step test already carries the per-transition property.

## Renamed rather than duplicated

Taking the second option, because the test does prove something the 40-step one doesn't: five 300-byte frames are offered to a **2-frame / 700-byte** window, so admits are **refused mid-run**, and conservation must survive that. A refusal that charged anything, or a settle-after-refusal releasing something never charged, surfaces as a window that doesn't return to its starting state.

Duplicating the per-transition assertions here would have added a third copy of a property already covered, while leaving the refusal path — which nothing else exercises — still unnamed.

## Added the non-vacuity check it never had

The old "NOT VACUOUS" comment defended assertions that *were* vacuous. The real risk is that the run never pressures the window at all, making it the clean round-trip test under another name. So it now asserts refusals actually occurred.

**Mutation-verified:** widening the window so nothing is refused fails it with *"no admit or settle was refused; the window was never pressured"*.

Also dropped a dead third element from the settle op tuples.

## Verification

784 tests, 0 failures. Credo-clean, format-clean, `openspec validate --strict` valid, `git diff --check` clean.

## Still open, unchanged and not claimed

PubAck correlation (task 3.5 + 3.3(c) publisher integration), encoded-size binding (task 3.4), deadline policy.


mfreeman451 (commented):
Reviewed exact head `f6ecd32372b157136844f031db36dd536f938fdf` against base `507f3500fb54703a7305f3fc3f8056196439b131`.

The previous final-state-only overclaim is closed: this test no longer claims to prove the bound after every transition, and the renamed round-trip/conservation property is useful because the schedule pressures the window. One P2 remains in the new non-vacuity evidence:

**[P2] Prove that an admit was refused, rather than pooling every operation's errors.** `publish_window_test.exs:551-581` increments the same `refusals` counter for failed admits and failed settles. That lets the assertion pass with zero refused admits: mutate the frame-capacity branch in `admit/4` from `{:error, :frame_credits_exhausted}` to `{:ok, w}` (false success without recording or charging the frame), and the later `settle(3)` supplies the sole refusal while `final === start` still holds. I ran that exact mutant against this test at line 542; it passed 1/1. The widening-window control does not exercise this opposite failure direction. Track admit and settle refusals separately, or retain an exact result trace, and assert that the pressured admit is refused with `:frame_credits_exhausted` (and ideally that its later settle is `:not_outstanding`).

Independent verification on the unmodified head: the focused file passes 31/31; formatting, strict OpenSpec validation, and `git diff --check` pass; all exact-head PR checks, including BazelCI, are green. No production/runtime blocker found.

PR #4020, opened 2026-08-25
fix(discovery): cascade mapper job deletion to its seeds and UniFi controllers

mfreeman451:
Deleting a discovery job silently orphaned everything that referenced it.

`mapper_job_seeds.mapper_job_id` and `mapper_unifi_controllers.mapper_job_id` had **no foreign key at all** — only `credential_secret_id` did. So removing a job left its seeds and controllers pointing at an id that no longer exists: **invisible in the UI** (which lists jobs), still present in the database, and unreachable for an operator through the product.

## Observed, not hypothetical

A job deleted on **2026-08-11** while splitting two networks left behind:

```
orphaned seed:       192.168.1.1  -> job 340c1db1-...  (gone)
orphaned controller: farm01       -> job 340c1db1-...  (gone)
```

The links that job produced — **390 of 521, 75% of the topology** — froze on that date and were still being drawn as current two weeks later.

## Fix

A foreign key with `ON DELETE CASCADE`, so this is the database's problem rather than something application code has to remember. **CASCADE rather than SET NULL** because neither row means anything without its job: a seed with no job is never scanned, a controller with no job is never polled.

## Proven against a real database

```
before_delete seeds=1 controllers=1
after_delete  seeds=0 controllers=0
```

And creating an orphan is now impossible:

```
ERROR: insert or update on table "mapper_job_seeds" violates foreign key
constraint "mapper_job_seeds_mapper_job_id_fkey"
```

Both constraints confirmed as `CASCADE` in `information_schema`.

## On the migration's DELETE

It carries `serviceradar:allow-startup-maintenance` with justification, and the baseline gate accepts it. The reasoning: **adding the constraint fails while violating rows exist**, so removing them is a precondition of the schema change rather than maintenance done alongside it. It's bounded to orphans in two small operator-managed config tables (single digits on the affected deployment) and restricted to rows whose referenced job doesn't exist — rows a foreign key would have made impossible to create.

`helm` `expectedVersion` moves with it so `wait-migrations` still proves the schema is current.

## Relationship to #4019

Complementary, not overlapping. #4019 stops **stale edges** being asserted as current; this stops the **dangling config rows** that produced them from being created in the first place. Together: the job's evidence ages out, and its dependants leave with it.


mfreeman451 (commented):
lgtm

PR #4024, opened 2026-08-25
test(edge): assert the exact result trace instead of pooling refusals

mfreeman451:
The last P2 against `f6ecd323` (merged as #4018 before this was ready). Reconciled onto current `usp-01-proposal`, 0 behind.

## The pooled counter could be satisfied by the wrong operation

You're right, and it's the direction my control never tested. A single `refusals` counter incremented for both failed admits and failed settles, so:

- make `admit/4`'s frame-capacity branch return `{:ok, w}` without recording or charging
- **no admit is refused** — the later `settle(3)` of that never-admitted sequence supplies the only failure
- `final === start` still holds, because nothing was ever charged

The widening-window control only proved *too few* refusals; it could never catch a refusal **attributed to the wrong operation**.

## Exact ordered trace instead

Every operation and its outcome is now asserted in order, pinning the two failures to specific positions for specific reasons — sequence 3 cannot be admitted because the **frame** grant is full, and therefore cannot later be settled:

```elixir
{:admit, 3, :frame_credits_exhausted},
...
{:settle, 3, :not_outstanding},
```

## Verification against your exact mutant

**KILLED**, reporting `{:admit, 3, :ok}` where `{:admit, 3, :frame_credits_exhausted}` was expected. The byte-capacity equivalent (`{:ok, w}` on the byte branch) is killed too.

784 tests, 0 failures. Credo-clean, format-clean, `openspec validate --strict` valid, `git diff --check` clean.

## Still open, unchanged and not claimed

PubAck correlation (task 3.5 + 3.3(c) publisher integration), encoded-size binding (task 3.4), deadline policy.


mfreeman451 (commented):
Reviewed exact head `7606f8ba411df2508af36249454734285422f13d` against base `ba3962bf76e89a766632688c63ce6c8bb3f40db3`.

No findings. The exact ordered trace closes the pooled-refusal hole: it independently requires the third admit to fail at its own position with `:frame_credits_exhausted`, and requires the later settlement of that unadmitted sequence to fail with `:not_outstanding`. The final whole-struct equality then carries the distinct conservation property.

I replayed both false-success mutations independently. Returning `{:ok, w}` from the frame-capacity branch fails this test with `{:admit, 3, :ok}` at the third trace position. Returning `{:ok, w}` from the byte-capacity branch fails the dedicated byte-bound test. Both source mutations were restored cleanly.

The diff remains test-only and does not claim the still-open PubAck-correlation, encoded-size, or deadline-policy work. Focused verification passes 31/31; formatting, strict OpenSpec validation, and `git diff --check` pass; all exact-head PR checks, including BazelCI, are green. Part 2 has no open review finding at this SHA.

Issue #4022, opened 2026-08-25
Link-local addresses must not be identity alias evidence


mfreeman451:
Found while investigating why `192.168.1.1` had not merged into the farm01 chassis (fixed in #4009/#4012). Not a regression from that work — a pre-existing hazard the chassis-merge investigation surfaced.

## The problem

IPv6 link-local addresses (`fe80::/10`) are **unique per link, not globally**. Two unrelated devices on two different L2 segments routinely carry the same `fe80::` address — it is commonly derived from the MAC via EUI-64, but a device with the same NIC vendor, or one using a stable-privacy or manually configured address, can collide outright. They are also the *most* commonly observed address on an SNMP-discovered interface, so they end up as high-volume alias evidence.

When they are recorded as identity aliases, "these two devices share an address" stops being merge evidence and starts being noise.

## Measured

When examining bad merge candidates during the chassis-split investigation, **4 of 5 keyed on an `fe80::` address**. The chassis merge landed and behaved correctly (`merge_audit` 23→24, `chassis_alive` 2→1, live devices 126→125, no over-merge), so this did not fire in that path — but the candidate pool it draws from is contaminated.

Related, already fixed: #3905 dropped link-local *primaries* 22→9 (they were climbing 18→22 before the fix). That addressed link-local as a **primary IP**. This issue is link-local as **alias evidence**, which is a different code path.

## Where the gate belongs

`Identity.AliasPolicy` — and it must be called by **both** producers, not just one. The recurring failure mode in this area (see the mapper IPv6/alias-gate work) is a policy that lives in one producer and is silently bypassed by the other. `grep` for the attribute, not for the call site.

`Identity.Address` (added in #4009) already classifies and ranks addresses including link-local, so the classification exists; it just isn't consulted here.

## Acceptance

- [ ] `fe80::/10` never becomes alias evidence, from either producer
- [ ] Test proves it for **both** producers independently — a single shared-helper test does not prove the second caller uses the helper
- [ ] Existing link-local alias rows: **archive, do not mark stale.** Alias debris in this system self-revives (see the netprobe passive over-merge arc, #3971/#3976/#3979)
- [ ] Re-count merge candidates keyed on link-local after the fix — gate on the artefact, not on the job reporting success

## Not in scope

Link-local as a *primary* IP (done, #3905) and link-local in `ip_addresses` generally — an interface legitimately has one, and recording it is correct. This is only about treating it as cross-device identity evidence.

[...]

PR #4027, opened 2026-08-25
fix(inventory): current-state interfaces and no link-local alias evidence

### LINKED ISSUE #4021: Re-key discovered_interfaces to current state (the 1M-device scale fix)
Remaining work from `openspec/changes/refactor-interface-observation-persistence`. The prerequisites landed in #4012; this is the change itself.

## The problem, measured

`platform.discovered_interfaces` is append-only and unbounded. On a **4-device** sample:

| property | value |
|---|---|
| rows | 36,816 |
| distinct semantic states | **374** |
| amplification | **98×** |
| growth | 767 rows / device / day |
| hypertable | no |
| retention | none |
| compression | none |

At the target deployment size (50k–1M+ devices) that's ~38M–767M rows/day, ~99% of it restating what's already stored.

## What must change, and it has to be ONE commit

- `identity :unique_interface, [:timestamp, :device_id, :interface_uid]` → drop `:timestamp`. That identity **is** the append-only mechanism.
- `reassignments.ex:219-235` collision probe re-keys to match.
- The JSON:API id at `interface.ex:69` embeds the timestamp — external contract.

**These cannot be staged separately.** The probe decides move-vs-destroy: non-colliding rows are moved to the merge survivor, colliding ones destroyed. Re-keying the probe while the identity still contains `:timestamp` would classify rows as colliding that today don't, and **destroy rows that today are legitimately moved**. And it runs inside `Ash.transact` with `rollback_on_error?: true` (`merge_engine.ex:226`), so a mismatch rolls back **every device merge** — presenting as a merge bug, not an interface bug.

## The key is `(device_id, interface_uid)`, not `if_index`

`if_index` is `allow_nil? true` and `inventory/sync/interfaces.ex` never sets it, so sync rows carry NULL and a unique index wouldn't dedupe them. `interface_uid` is `allow_nil? false` and already `primary_key? true`.

On farm01 both keys look identical — 340 distinct pairs each, zero NULLs — but only because every interface there comes from the mapper. Exactly the coincidence that makes a wrong key look right until another producer appears.

## Already done (#4012)

- [x] `upsert_fields` enumerated at both writers, deliberately asymmetric — sync never sets `if_index`/`if_speed`/`speed_bps`/`if_admin_status`/`if_oper_status`/`if_type`/`mtu`/`duplex`/`available_metrics`, so a shared list would blank all nine on every sync run once conflicts start happening
- [x] canonical `ip_addresses` ordering (also fixed a live `primary_ip` bug)
- [x] reader inventory — 19 Elixir modules + SRQL classified

## Remaining checklist

- [ ] Classify the four readers named as **unclassified rather than assumed safe**: `rust/srql/src/query/interfaces/sql.rs`, `.../interfaces/stats.rs:35`, `.../logs/metadata.rs:123`, `web-ng .../live/interface_live/index.ex:290` (`stats:count() as total`, likely a second user-visible count)
- [ ] Pin the targeting count's device-semantics in a web-ng test (a device with several matching interfaces counts once)
- [ ] Keep `timestamp` meaning "last observed", bumped every poll — `mapper_results_ingestor.ex:2193` (`ago(6,"hour")` liveness) and `interface_data.ex:661/:720` (`time:last_3d`) both depend on it
- [ ] The rekey commit itself
- [ ] Change-only history store — **not required day one**: no reader anywhere needs multiple rows over time. Wanted for causal outage analysis ("what changed on the network around the time this broke"), and it must record *what changed* (previous value, new value, which fields), not snapshots
- [ ] Verify: row count stops growing while nothing changes (task 6.3), **and** a genuine change is still captured (6.4) — a detector that suppresses everything passes 6.3 perfectly

## Also unknown

Whether any dashboard or alert threshold is calibrated on the inflated `:active` interface count.

mfreeman451:
## Describe your changes

Closes #4021 and #4022.

**#4021 — Re-key `discovered_interfaces` to current state.** The identity is now `(device_id, interface_uid)` so a poll upserts instead of appending. `timestamp` remains last-observed. The merge collision probe in `reassignments.ex` and the JSON:API id move in this same commit, because they have to stay in lockstep with the identity. The Timescale hypertable is converted to a regular table (a unique key without `timestamp` is illegal on a hypertable). Historical restatements are discarded; that is stated in the migration. A change-only history store is still not day one.

**#4022 — Link-local addresses must not be identity alias evidence.** `fe80::/10` and `169.254/16` are rejected as `:ip` aliases from both producers (mapper `alias_ips_for_role/3` and `AliasEvents.process_and_persist/2`), and `DeviceAliasState.detect` enforces the same policy so a forgotten caller cannot persist them. Existing link-local alias rows are archived by a new DIRE step (`link-local-alias-archive`), not marked stale — stale aliases self-revive.

- https://github.com/carverauto/serviceradar/issues/4021
- https://github.com/carverauto/serviceradar/issues/4022

- [x] I have signed the DCO?

## Tests run locally

- `mix test --no-start` on focused core unit tests: 77 tests, 0 failures (identity keys, mapper alias_ips_for_role, AliasPolicy, DIRE step order, sync upsert_fields).
- Integration tests against a disposable `srql-fixtures` scratch database:
  - #4022: 7 tests, 0 failures (`AliasEventsLinkLocalTest`, `LinkLocalAliasArchiveTest`).
  - #4021: migration `20260825020000` applied; `InterfaceCurrentStateTest` 2 tests, 0 failures (upsert not append; three interfaces count as one SNMP target).
- Not run: full `make test` / Bazel `//...` unit sweep (long/remote). Live farm/demo growth check (OpenSpec 6.3/6.4) and post-deploy re-count of `fe80::` merge candidates are still operator verification after rollout.


mfreeman451:
## Summary

Independent grok CLI review of https://github.com/carverauto/serviceradar/pull/4027 (`fix/iface-current-state-and-link-local-alias` vs `staging`). I did not author this change.

The direction is right: `#4021` needs `(device_id, interface_uid)` as the current-state key with `timestamp` as last-observed, and `#4022` needs `fe80::/10` / `169.254/16` out of identity-alias evidence. Mapper `alias_ips_for_role/3`, `AliasEvents.process_and_persist/2`, and `DeviceAliasState.detect` are actually wired to `AliasPolicy`, and the merge collision probe / JSON:API id / Ash identity were moved together on purpose.

It is not shippable as-is. The Ash resource primary key was not moved with the identity, the hypertable collapse can throw away mapper operational columns the live upsert path was carefully taught not to blank, and existing `fe80::` alias rows remain live merge evidence until an operator happens to run DIRE. `#4022` measured 4/5 bad merge candidates keyed on link-local; that path is the reader, not the writer.

## Must-fix

### 1. Ash `Interface` primary key still includes `timestamp` (`elixir/serviceradar_core/lib/serviceradar/inventory/interface.ex`)

`#4021` is explicit that identity, the merge collision probe, and the JSON:API id have to move in lockstep. This PR moves:

- `identity :unique_interface` → `[:device_id, :interface_uid]`
- JSON:API `primary_key keys` → `[:device_id, :interface_uid]`
- `Reassignments.fetch_existing_interface_keys/3` → uid-only

It does **not** move the Ash resource key:

```elixir
# Composite primary key: timestamp + device_id + interface_uid
attribute :timestamp, :utc_datetime do
  allow_nil? false
  primary_key? true
```

Postgres after the migration is `PRIMARY KEY (device_id, interface_uid)`. Ash still believes a row is `(timestamp, device_id, interface_uid)`.

That is the same class of mismatch `#4021` warned about for the probe. `Ash.bulk_update` / `Ash.bulk_destroy` during merge identify records by the resource primary key, so a concurrent mapper/sync upsert that bumps last-observed `timestamp` makes `WHERE timestamp = $old AND device_id = $d AND interface_uid = $u` miss. The surviving device then either keeps a duplicate uid (PK violation, merge rolls back — `rollback_on_error?: true`) or the losing row is stranded.

`InterfaceCurrentStateIdentityTest` only asserts `identity.keys`. It cannot fail this bug. Pin `Ash.Resource.Info.primary_key(Interface) == [:device_id, :interface_uid]` (and `refute :timestamp in ...`) in the same test, and drop `primary_key? true` on `:timestamp`.

### 2. `DISTINCT ON` latest-row-wins can discard mapper SNMP columns (`elixir/serviceradar_core/priv/repo/migrations/20260825020000_rekey_discovered_interfaces_current_state.exs`)

```sql
INSERT INTO ...discovered_interfaces_current
SELECT DISTINCT ON (device_id, interface_uid) *
FROM ...discovered_interfaces
ORDER BY device_id, interface_uid, timestamp DESC, created_at DESC NULLS LAST
```

Historical restatements being dropped is stated and acceptable. Copying the **newest row wholesale** is not the same thing.

The live writers already know this hazard: `inventory/sync/interfaces.ex` refuses to put `if_index`, `if_speed`, `speed_bps`, `if_admin_status`, `if_oper_status`, `if_type`, `mtu`, `duplex`, `available_metrics` in `upsert_fields`, because a sparse sync observation would NULL them out once `:unique_interface` is current-state. That discipline applies only *after* the migration. The collapse runs against append-only history, where a later sync row is a full replacement.

If the latest observation for a key is a sync row, current state loses `if_index`. SRQL error-metric joins are `tm.if_index = di.if_index`. Interfaces that the mapper does not re-walk after deploy never get those columns back.

Fix: per key, keep the latest `timestamp` / metadata / last-observed fields, and take latest-non-null for the nine mapper-only operational columns (the same list already named in `InterfacesUpsertFieldsTest`). Prove it with a migration test that inserts a rich mapper row then a later sparse sync row and asserts `if_index` survives.

### 3. Identity *readers* still treat leftover `:ip` aliases as merge evidence

Writer side is gated. Consumer side is not.

`AliasGuard.maybe_merge_ip_alias_device/3`, `Resolver.lookup_alias_device_id/4`, and `DeviceLookup.read_alias_states/3` all query `alias_type == :ip` with no `AliasPolicy` check. `MapperResultsIngestor.find_device_uid_by_alias/3` still returns non-archived/non-replaced `:ip` rows, including `:stale`, and `maybe_reactivate_alias/2` still revives stale.

`#4022` measured **4 of 5** bad merge candidates keyed on `fe80::`. DIRE `link-local-alias-archive` is the cleanup, but it is operator-invoked, default-dormant (`mix serviceradar.dire_remediation`), limited to 5_000 rows per run, and sits *after* `agent-links` / `proxmox-dups` in `@step_order`. Until someone runs execute, ingest-time `ip_alias_conflict` keeps consuming the leftover rows.

`#4022` acceptance is `fe80::/10` never becomes alias evidence, and “grep for the attribute, not the call site.” The attribute is still read for merges.

Do both:

- Fail closed in the readers: if `not AliasPolicy.valid_alias_ip?(ip)`, do not look it up, do not merge on it, do not reactivate it. Same predicate as the writers.
- Archive existing rows in a migration (or an execute-on-deploy step), not only in DIRE. DIRE can stay as the replayable sweeper. Do not wait on a mix task to make `#4022` true in production.

Also extend the archive query to `:interface_ip` (the resource validation now rejects those too) and strip leftover `ip_alias:fe80::…` metadata keys the way `netprobe-alias-debris` does. Persist is gated; metadata is still sitting there for the next caller that forgets the policy.

## Should-fix

- **`AliasPolicy` reimplements classification instead of calling `Identity.Address`.** `#4022` says Address already classifies this. `Address.classify/1` unwraps `%zone`, `/cidr`, and IPv4-mapped `::ffff:a.b.c.d`. `AliasPolicy.valid_alias_ip?/1` does not. `::ffff:169.254.1.1` and `::ffff:127.0.0.1` are accepted as identity aliases. Drive the predicate off `Address.classify/1` (reject `:link_local`, `:loopback`, `:unspecified`, `:invalid`) and add those two mapped cases to `AliasPolicyTest`.
- **NDP / census / fingerprint / DPI / process decoders still write `ip_alias:fe80::…` into device metadata** (`census.ex:140-144` and siblings). `Sync.Normalize.enrich_alias_metadata/1` also appends `update.ip` with no policy check. Persist is filtered, so this is not a second writer of *rows*, but it is a missed call site for the attribute the issue told you to grep. Filter at the metadata source too.
- **`AliasEvents.alias_values/1` comment is false for `:collector_ip`.** The comment says `current_ip` *and* `collector_ip` from `base` go through the policy. The filter only matches `{:ip, _}` and `{:interface_ip, _}`. `{:collector_ip, _}` falls through to `{_type, _} -> true`. Not a merge key today, but it is an address-typed alias the resource validation also does not cover (`@address_types [:ip, :interface_ip]`).
- **`rust/srql/src/schema.rs` still declares `discovered_interfaces (timestamp, device_id, interface_uid)`.** SRQL is read-only so this is not an insert bug, but it is the same identity the Elixir side just changed. Update the Diesel table key to match.
- **Index coverage after `DROP TABLE … CASCADE`.** Recreated: PK, device FK, `idx_discovered_interfaces_device`, `(device_id, if_index)`, metadata GIN. Not recreated: `discovered_interfaces_timestamp_idx`, `idx_discovered_interfaces_device_time`, `idx_discovered_interfaces_device_uid_time`, `idx_discovered_interfaces_device_if_index_time`. SRQL time-range filters are `di.timestamp >= $1 AND di.timestamp <= $1`. After rekey that is a sequential scan on last-observed. Add a `timestamp DESC` index if `time:last_3d` stays on the Interfaces index.
- **DIRE step order / batching.** Put `link-local-alias-archive` before any step that can merge (`agent-links`, `proxmox-dups`). The 5_000 `LIMIT` plus `would_archive == 0` as an idempotence check is wrong the moment a fleet has more than 5_000 leftover rows, or the moment two execute runs are needed.
- **Tests that cannot catch the bugs above.**
  - `InterfaceCurrentStateIdentityTest` does not inspect Ash `primary_key`.
  - `LinkLocalAliasArchiveTest` asserts `would_archive >= 2` / `archived_alias_states >= 2` / second-run `would_archive == 0` against the *whole table*. Pin the seeded ids. `>= 2` passes on unrelated fixture rows; `== 0` fails on any leftover `fe80::` in a shared scratch DB.
  - `AliasEventsLinkLocalTest` never sends `169.254/16`. Mapper tests do; the “both producers independently” bar in `#4022` is not met for APIPA on the AliasEvents path.
  - SNMP targeting: the behavioural assert lives in core (`Ash.Query.distinct(:device_id)` in the test), and web-ng only pins the source string. That is explained, but deleting `distinct` in `targeting.ex` and rewriting it as a subquery would still satisfy the regex.

## Nits

- Stale comment on `Interface` attributes: “Composite primary key: timestamp + device_id + interface_uid”.
- `timescaledb_pkey_violations?/1` matching `\d+_\d+_\w+_pkey` is dead for interfaces once this is a regular table. A unique violation will now surface as a real bulk error (good); the Timescale skip path can be dropped or narrowed to topology_links.
- `Address` moduledoc now says link-local must never become a device alias, but `Address` itself still only ranks primaries. Easy to read as if `Address.rank/1 == 20` were the alias gate. It is not; `AliasPolicy` is.
- Migration `@disable_ddl_transaction false` is the Ecto default. The `INSERT … SELECT DISTINCT ON` of an unreaped table is a long exclusive lock; worth a comment about expected row count / that 3-day retention may not exist on every deploy (`#4021`’s 4-device sample was “hypertable: no, retention: none”).
- JSON:API ids lose the timestamp segment. That is the requested lockstep, but it is a breaking external contract for any client holding `ts~device~uid` ids. Call it out in the PR body.
- `create_detected` is the only action with `AliasValue`. `record_sighting` / `reactivate` / `confirm_from_sweep` will still advance an existing bad row if a reader looks it up. Once readers fail closed (must-fix 3) this is moot.

---

**Verdict:** do not merge. Fix the Ash primary-key lockstep, the migration column-coalesce, and the identity-reader / leftover-row hole. The producer tests are the right shape; they do not cover the three holes above.

grok CLI review

mfreeman451:
## farm01 roll (PR 4027)

Mike approved this roll. No merge, no gitops commit.

### Images
- Rolled tag: `sha-2fdd94bddb1f0cd965778af6120ab8c482dedd30` (`2fdd94bd`)
- Rebuilt: `serviceradar-core-elx@sha256:e7df8dcfd2f6c4af2b47599951f4c5650990c4352901b8eaabd86d3a0bab1cdd`
- Copied forward from the previous farm01 tag: datasvc, flow-collector, log-collector, rperf-client, tools, trapd, trivy-sidecar
- Left on existing digest pins (not rebuilt): agent, agent-gateway, **web-ng** (targeting-count change is test-only)

First attempt used `sha-7e75605fbc`. The pre-upgrade migrations job failed: `array_agg(available_metrics jsonb[])[1]` is `jsonb`, not `jsonb[]` (`datatype_mismatch`). Helm rolled back to revision **122**. Fix is `2fdd94bd`; collapse SQL now round-trips `available_metrics` through text. Proved on farm01 CNPG temp tables and the core integration test before the second roll.

Live farm01 pins `image.digests.core`, which short-circuits `global.imageTag`. The upgrade therefore set both:

```
--reuse-values
--set global.imageTag=sha-2fdd94bddb1f0cd965778af6120ab8c482dedd30
--set image.digests.core=sha256:e7df8dcfd2f6c4af2b47599951f4c5650990c4352901b8eaabd86d3a0bab1cdd
```

Did **not** `helm -f` gitops `clusters/farm01/serviceradar/values.yaml`.

### Helm
- chart `1.4.42`, ns `serviceradar`, context `farm01`
- revision **125** (failed 123 → rollback 124 → success 125)
- `deploy/serviceradar-core` 3/3 on the new digest

### CNPG (ns `serviceradar`)
- `platform.schema_migrations` has **20260825030000** and **20260825031000**
- `discovered_interfaces` PK is **`(device_id, interface_uid)`** (not timestamp); not a hypertable
- OpenSpec **6.3**: `COUNT(*)` = **307** then **307** ~79s later (07:20:18Z / 07:21:37Z). Stable.
- OpenSpec **6.4**: did not take an interface down. Query to re-check a genuine change:

```sql
SELECT device_id, interface_uid, timestamp, if_index, if_oper_status
FROM platform.discovered_interfaces
WHERE device_id = '<uid>'
ORDER BY interface_uid;
```

After a real oper-status / address change: same `(device_id, interface_uid)` row, newer `timestamp`, updated operational columns. A new row for that key would mean the rekey did not stick.

- **#4022 leftover merge candidates** (`alias_type IN ('ip','interface_ip')`, not archived, `sr_address_rank=20`): **0**. 33 rows archived. 11 devices still have `ip_alias:fe80::…` **metadata** keys (census/NDP writers are still unfiltered; persist path is gated).

### alma-test01
tmux `0.0` idle `[root@alma-test01 mfreeman]#`. `systemctl status serviceradar-agent --no-pager | head`: **active (running)** since Mon 2026-08-24 22:02:03 CDT. No agent rebuild/scp.

PR #4053, opened 2026-08-26
Refresh the integration lane snapshot after new device-type tests

mfreeman451:
## Why

`//build:integration_shards_topology_test` fails on `staging` itself, so every branch cut from it
inherits a red BazelCI. Reproduced on plain `a215af2695`:

```
Expected "[26, 22, 22, 23, 22, 22, 23]", but got "[26, 22, 21, 22, 23, 23, 23]"
Expected "[187, 192, 192, 191, 190, 190, 190]", but got "[188, 193, 192, 191, 191, 191, 191]"
Expected "[213, 214, 214, 214, 212, 212, 213]", but got "[214, 215, 213, 213, 214, 214, 214]"
```

## What happened

Nothing is wrong with the partition. `7b0d898961` (#4046, "an integration must not overwrite a
hand-set device type") added ~200 lines of cases to
`elixir/serviceradar_core/test/serviceradar/inventory/sync_ingestor_vendor_type_test.exs` -- an
**existing** file. The source-count assertions in the very same test are untouched and still pass:
126 async sources, 160 serial sources, 286 selected. Five more test *cases* inside one existing
file changed what the greedy bin-packer emits, which moved all three lane snapshots.

Balance is unchanged: sizes still sum to 160, and the spread between the fullest and emptiest lane
stays at 5 tests (193 vs 188, previously 192 vs 187).

## The change

Refresh the three snapshots to what the partitioner now produces. They stay mutually consistent --
`weights == test_counts + sizes` for every lane:

| lane | size | tests | weight |
| ---- | ---- | ----- | ------ |
| serial_0 | 26 | 188 | 214 |
| serial_1 | 22 | 193 | 215 |
| serial_2 | 21 | 192 | 213 |
| serial_3 | 22 | 191 | 213 |
| serial_4 | 23 | 191 | 214 |
| serial_5 | 23 | 191 | 214 |
| serial_6 | 23 | 191 | 214 |

These arrays are load-bearing -- they are what stops a lane silently growing past its connection
budget -- so they have to be refreshed deliberately whenever the case count moves, rather than
loosened. That deliberate refresh is what #4046 missed.

## Verification

- `//build:integration_shards_topology_test` passes.
- `//build/...` green (21 tests).
- `make test` green: **199 test targets, 199 pass**.


mfreeman451:
Superseded by https://github.com/carverauto/serviceradar/pull/4056.

This branch refreshed the lane snapshots against a 4046-only tree. #4049 and #4052 landed on staging afterwards, so the arrays here no longer match the packer (and BazelCI is red on the merge). Closing in favor of 4056, which takes the lists from current staging.

PR #4056, opened 2026-08-26
fix(test): refresh serial lane snapshots after 4046/4049

mfreeman451:

Signed-off-by is on `e466d8c643`.

## Describe your changes

`//build:integration_shards_topology_test` is red on `staging`. Extra cases landed in **existing** serial files (#4046 vendor-type, #4049 isolation ingest/coverage). Source counts are unchanged (126 async, 160 serial, 286 selected). The greedy packer moved.

#4053 refreshed the arrays against a 4046-only tree. #4049 and #4052 merged after that, so those numbers are already wrong. This PR takes the three lists from current `staging`.

| lane | size | tests | weight |
| --- | --- | --- | --- |
| serial_0 | 26 | 188 | 214 |
| serial_1 | 22 | 193 | 215 |
| serial_2 | 22 | 193 | 215 |
| serial_3 | 23 | 192 | 215 |
| serial_4 | 22 | 191 | 213 |
| serial_5 | 22 | 191 | 213 |
| serial_6 | 23 | 191 | 214 |

`weights == tests + sizes` on every lane. Spread is still 5 tests (193 vs 188). Sizes still sum to 160.

Also registers `source_policy_self_report_test.exs` as `load_only` so the exhaustive TSV inventory matches `ALL_TEST_SRCS` after #4052. Selected count stays 286; load_only 508 → 509.

Supersedes #4053.

None. Follow-up to #4046 / #4049 / #4052. Closes the gap #4053 tried to fill.

- [x] I have signed the DCO?

Verified:

- `//build:integration_shards_topology_test` PASS
- `//:ci_heavy_gate_contract_test` PASS
- `python3` exhaustive inventory + lane projection tests PASS


mfreeman451:
Closing as obsolete — **#4054** (merged) deletes the three arrays this PR refreshes, so there is nothing left to keep in sync.

Your description made the case better than mine did:

> #4053 refreshed the arrays against a 4046-only tree. #4049 and #4052 merged after that, so those numbers are already wrong.

That is the treadmill in one sentence: #4053 did this once and was stale before it landed, and this PR is the second attempt. The numbers were bin-packer output, not a contract anyone chose, so they went stale on contact with any other merge — and because BazelCI tests the **merge** of a branch with its base, whoever ran second inherited a red build caused entirely by whoever ran first.

It also sharpened the diagnosis in a way I had understated: you noted the extra cases landed in **existing** serial files. So it is not only adding a test *file* that invalidates the arrays — adding a test *case* to an existing file does it too. The treadmill is tighter than I had described.

#4054 replaces all of it with the property that actually matters. The lanes run in parallel, so what needs protecting is that no lane is much heavier than the others — now asserted as `max <= mean * 1.25` and `min >= mean * 0.75`, plus every lane carrying work. The bounds sit far outside the observed spread, so ordinary churn cannot trip them, and when the assertion does fire it prints the offending distribution instead of "Expected X, got Y".

The same defect existed twice more and is fixed in the same PR:
- the three absolute source totals in the same test — `126 + 160 == 286`, so the only invariant was that the two projections partition the selection exactly
- `ci_heavy_gate_contract_test.py` — `286 + 508 == 794`, same shape, sitting directly above the set-equality check that was already doing the real work

Both are relations now, with floors as truncation tripwires.

**One thing this PR was not responsible for and could not fix:** `integration_tests_serial_0` and `integration_tests_serial_3` fail at Exit 2 on this branch AND on #4054, which share none of each other's changes. That is a pre-existing staging breakage — 2 of 7 lanes, Exit 2 being a harness/setup failure rather than a test assertion. It is now the actual blocker on the queue and is being looked at separately.

Thanks for the lane snapshot table in the description — it was the clearest evidence available that the numbers could not be maintained by hand.

PR #4058, opened 2026-08-26
fix(inventory): scope active-IP conflict resolution and identity cache by partition

mfreeman451:
## What

Three related fixes so that partition, not IP alone, decides which device a
write lands on. The live unique index has been `UNIQUE (partition, ip)` since
the isolation-partition migration; several code and test paths still assumed
`UNIQUE (ip)`.

1. **`inventory/sync/device_writes.ex`** — the active-IP conflict flow was keyed
   on `ip`, so a monitoring and an isolation copy of the same address collided
   with each other. Keyed on `{partition, ip}` throughout.

2. **`identity/device_lookup.ex`** — `IdentityCache` is keyed on IP alone, so a
   cached record from another partition was returned as an answer, bypassing the
   partition-scoped DB query. A record whose partition differs is now treated as
   a miss, in both the batch and single-key paths.

   Deliberately **not** re-keyed on `{partition, ip}`: four invalidation sites
   (`device_notifier`, `device_identifier_notifier`, `device_alias_state_notifier`,
   `sync/state_events`) call `IdentityCache.delete/1` with a bare IP, so changing
   the key format would silently turn every one of them into a no-op and strand
   stale entries forever — a worse bug than this one.

   **This one is latent:** every caller passes `use_cache: false` today, so no
   current path could hit it. It is a guard for the opt-in the moduledoc invites.

3. **`sweep_results_flow_e2e_test.exs`** — stale expectations, not product bugs.

## The test failures were the tests, not the code

`SweepGroup.partition` is the device-lookup partition (890854c87b). These tests
paired devices created in `"default"` with invented per-test sweep group
partitions — harmless when partition was cosmetic, but once ingest scoped its
lookup by it, the lookup matched nothing and ingest created a **duplicate
device**. That surfaced three different ways: `devices_created == 1`, a
`BadMapError` from `get_by_ip` returning two rows, and availability never being
written.

The isolation test additionally asserted unavailability after **one** failed
sweep, but marking a device unavailable deliberately requires
`@unavailable_threshold` (2) consecutive failures. It now sweeps twice, so it
asserts partition isolation rather than tripping over hysteresis.

Verified directly that the product behaviour is correct: with two failed sweeps
the isolation copy goes `is_available=false` while the monitoring copy stays
`true` and is never touched (`sweep_consecutive_failures` stays `nil`).

## Verification

Run against the `srql-fixtures` scratch DB (these files are `@moduletag
:integration`, which `make test` filters out, so normal CI does not run them):

| | tests | failures |
|---|---|---|
| baseline | 58 | 11 |
| after | 58 | **1** |

Zero newly-broken tests (`comm` diff of failure names, both directions).

The one remaining failure — `one integration still reclassifies a type another
integration inferred` (expects `Switch`, gets `Tablet`) — is **pre-existing and
unrelated**. Proven by stashing the library changes and re-running it alone: it
still fails. Filed separately rather than fixed here.

Gates: `mix format --check-formatted` clean, `mix credo --strict` no issues,
`MIX_ENV=test mix compile --warnings-as-errors` clean.


mfreeman451:
Stacked #4060 into this branch so one BazelCI run can demonstrate both lanes
green before anything merges.

The two blocking failures live in different shards, and each branch fixed only
one — so neither could ever show a green BazelCI on its own:

| run | sweep fix (#4058) | vendor fix (#4060) | serial_0 | serial_3 |
|---|---|---|---|---|
| #4058 | yes | no | **PASSED** | FAILED |
| #4060 | no | yes | FAILED | **PASSED** |

`sweep_results_flow_e2e_test.exs` is in `serial_0`; `sync_ingestor_vendor_type_test.exs`
is in `serial_3`. Lane membership comes from the deterministic partition in
`//build:integration_shards.bzl` and is not visible in the check name — a failed
shard surfaces only as `BazelCI FAILURE`, which is why this took two runs to see.

This merge is sequencing only; no new code. Both fixes verified present by content
on the merge commit.

PR #4079, opened 2026-08-27
feat(edge): separately bounded NATS publisher connections per lane (task 3.3(c) part 3b)

mfreeman451:
Completes task 3.3(c). The pools landed in #4038 with the connection half explicitly unclaimed:
the runtime had one connection (`:serviceradar_nats`) and `Connection.request/3` took no name, so
the accounting boundary and the transport boundary did not coincide.

## The publisher lane is NOT the wire traffic class

Keying the publishers on `EdgeRecordTrafficClass` looks obvious -- two of the three lanes are
named after its members -- and is wrong in both directions:

* The wire enum has **no recovery member**. It is `UNSPECIFIED | BULK | INTERACTIVE`, so a traffic
  class alone can never select the recovery publisher.
* Recovery is a **route profile** (`EDGE_RECORD_ROUTE_PROFILE_RECOVERY_CONTROL_V1`) and it pairs
  with **both** classes. `StreamRoute.active_lanes/0` lists four pairs; recovery is two of them.

So the key is the pair, collapsed onto three publishers:

| `{route_profile, traffic_class}` | publisher |
|---|---|
| `{durable_records, BULK}` | `:bulk` |
| `{durable_records, INTERACTIVE}` | `:interactive` |
| `{recovery_control, BULK}` | `:recovery` |
| `{recovery_control, INTERACTIVE}` | `:recovery` |

Recovery collapses because it resolves to one singular unpartitioned stream regardless of class
(`StreamRoute.resolve/1`). A second publisher there would buy no isolation and cost a connection.

## Why the shape is the isolation

The spec requires recovery to have "separate unborrowable storage, PubAck, and consumer capacity".
Because recovery is reachable **only** through a route profile, and the profile comes from the
effective control-plane grant rather than the frame, an agent cannot ask for recovery capacity.
Keyed on the wire class, any lane an agent could name would have been a lane it could saturate.

The count is bounded **by construction**, not by convention: `lanes/0` is a closed three-element
list, `connection_name/1` accepts nothing else, and `connection_names/0` takes no arguments -- so
no scope, agent, assignment, run, contract, package, or partition can allocate another. A test
asserts the arity, so the bound fails loudly if the function ever grows a parameter.

## `child_specs/2` is public on purpose

`Gnat.ConnectionSupervisor`'s default child id is the **module**, so four of them under one
supervisor collide on id and `Supervisor.init/2` starts only the first. The failure is silent in
the worst way: the app boots, the shared connection works, and three lanes have no connection at
all until something tries to publish. Making the spec list testable without a NATS server is what
lets a database-free test catch it; the connection-NAME assertions cannot, because the names are
correct either way.

## Verification

22 tests, 0 failures against an srql-fixtures scratch database. `mix compile
--warnings-as-errors` clean, `mix format --check-formatted` clean, `mix credo --strict` reports no
issues on the three touched modules. Four Bazel gates green on RBE: `ci_heavy_gate_contract_test`,
`integration_selection_equivalence_test`, `unit_tests_serviceradar_edge`,
`unit_tests_serviceradar_other`.

Mutation-verified -- each mutation confirmed to apply before running:

| mutation | result |
|---|---|
| recovery maps to `:bulk` | **killed** (2 failures) |
| fallback lane instead of refusal | **killed** (1) |
| one shared connection name for all lanes | **killed** (4) |
| `connection_names/0` drops the publishers | **killed** (1) |
| colliding child ids (Gnat's default) | **killed** (1) |

The first attempt at the last mutation left a trailing comma and failed to compile. That is a
failed experiment, not a kill, so it was redone with valid syntax and only the redone run counted.

## Still NOT claimed

This does not publish, does not settle from real PubAcks, and does not bind byte credits to
encoded frame size. Those remain owed by tasks 3.4 and 3.5, and `PublisherPool`'s moduledoc has
been updated to say so rather than left describing an absence this PR filled.


mfreeman451 (commented):
Reviewed exact head `42f3ea098ba67b61fb9293417411b2d9352af3e6` against base `c53e07113398930e75213fc4d2f0955a59ae6bc6`.

All exact-head checks are green, including BazelCI. I found two runtime blockers and three closure gaps that the current tests do not exercise.

### [P1] The lane connections and windows are not on the production publish path

`ServiceRadarAgentGateway.JetStreamPublisher` still calls `conn.request(route.subject, payload, opts)`. That is `ServiceRadar.NATS.Connection.request/3`, which hard-codes the shared `:serviceradar_nats` connection. The new named `request/4` has no production caller. Separately, no production supervisor starts a `PublisherPool`; only tests call `PublisherPool.start_link/1`.

Consequently, the three new Gnat connections are idle, every real edge publish still shares one Gnat mailbox, and no runtime process owns the lane-specific frame/byte window. This does not yet satisfy the task's requirement for separately bounded publisher connections/pools, and the claim that the accounting and transport boundaries now coincide is false at this head.

Either wire the fixed, internally-derived lane into the actual publisher and supervise exactly the three pool owners, or scope this PR honestly as topology preparation and leave task 3.3(c) open for that integration. This does not require claiming the still-deferred PubAck correlation or encoded-size binding.

### [P1] The tests never bind the declared topology to `Supervisor.init/1`

The connection test calls the pure `child_specs/2` helper and checks IDs, but it never checks what `init/1` actually returns or the registered name passed to each child. I reproduced two survivors independently:

1. Set every child's inner Gnat `name` to `:serviceradar_nats` while retaining the four unique child IDs: all 22 focused tests passed. At runtime the children then collide on the same registered name.
2. Change `init/1` to supervise only `[hd(child_specs(...))]`: all 22 focused tests passed. At runtime only the shared connection starts.

Bind the exact four-child inventory through `init/1` (or a live supervised instance), and assert for every child that its child ID and the registered Gnat name both equal the declared connection inventory.

### [P2] Recovery's accepted traffic-class surface is fail-open

The recovery clause accepts either current class by wildcard, and no negative covers `UNSPECIFIED` or a future/unknown class. Replacing the two intended recovery cases with one `_class` clause leaves all 8 `PublisherLane` tests green. Derive the declared traffic-class inventory and prove recovery accepts exactly BULK and INTERACTIVE, refusing every other declared value, with an explicit unknown-value control.

### [P2] `active_assignments/0` is checked against itself, not the normative mapping

The current test derives the expected lanes from `active_assignments/0`, so it proves only that the function is internally self-consistent. Mapping every active profile/class pair to `:bulk` leaves all 8 tests green. Bind the exact four assignments, or independently assert each assignment's result through the normative profile/class-to-lane rule.

### [P2] The claimed pool cardinality is not structurally bounded

`PublisherPool.start_link/1` still permits `name: nil` and arbitrary names, so callers can create any number of independent windows against the same lane connection. The closed `PublisherLane.lanes/0` list bounds declared connection names, but not aggregate pool capacity. Make the production supervisor own one fixed pool per lane and narrow the construction surface accordingly, or remove the structural-cardinality claim and state the actual ownership boundary.

Independent verification at this head: focused 22/22 tests, warnings-as-errors compilation, formatting, strict OpenSpec validation, and `git diff --check` all pass. The four mutations above were restored after measurement and the worktree is clean.

PR #4086, opened 2026-08-27
docs(openspec): composite_results stats and dashboard frame paging

mfreeman451:
## Describe your changes

OpenSpec proposal (`add-composite-results-stats-and-frame-paging`) for the long-term fix to fleet-sized composite-check dashboards. **Proposal only — no implementation in this PR.**

Today `in:composite_results stats:count() as n by verdict` is parsed and ignored, so every dashboard count is a client-side fold over a FrameRunner-clamped row sample (2,000 on current hosts). Raising that clamp just ships tens of MB of `inputs` jsonb over the dashboard channel every 15s.

This change specifies:

1. **`stats:` on `composite_results`** — `count()` grouped by `check` / `verdict` / `status`, plus `jsonb_each` unnest of `inputs` for vantage-point rollups. Unsupported aggregations **error** instead of silently dumping rows (**BREAKING** vs the current ignore; no in-tree caller).
2. **`api.srql.page(frameId, cursor)`** — FrameRunner already copies SRQL `pagination` onto frames and never passes a cursor back. Wire that through without rewriting the query or remounting the renderer.
3. **Device table pages of ≤200** so `in:devices uid:(…)` can name every row on the page (`MAX_FILTER_LIST_VALUES` is 200).

Customer packages (UAL `com.ual.armis.composite`) consume this in a follow-up repo after the host + SDK minor land. See `openspec/changes/add-composite-results-stats-and-frame-paging/`.

Validated: `openspec validate add-composite-results-stats-and-frame-paging --strict`.

N/A — follow-up to the Example Armis composite dashboard hitting the 2,000-row FrameRunner ceiling.

- [x] I have signed the DCO?

Docs-only change; `openspec validate --strict` passes. Implementation is gated on proposal approval (`openspec/AGENTS.md` Stage 2).


mfreeman451:
Parking the follow-ups that are not this PR so they cannot get dropped on merge:

- Host CHANGELOG + portal docs: #4099 (OpenSpec 5.1-5.2)
- SDK wrap + 0.2.0 minor: carverauto/serviceradar-sdk-dashboard#2 (OpenSpec 4.1-4.5)
- com.ual.armis.composite consumer: #4098 (OpenSpec 5.3). serviceradar-armis-dashboards is not on GitHub yet; wifi-dashboard is the network-map package, not this one.

PR #4091, opened 2026-08-28
fix(edge): make the lane windows gate the real publish path

mfreeman451:
Fix-forward for the review of `42f3ea09` / `a1189bc8`, opened against the current integration head.

**Repository state first:** #4079 merged at `42f3ea09`, the *pre-fix* head, so the earlier review
fixes in `a1189bc8` were orphaned -- on no PR and not on `usp-01-proposal`. They are rebased onto
`5a961d3c` here and carried forward with the new work.

## [P1] The windows now gate the real publish path

They were declared but inert: `publish_record/2` called `Connection.request/4` directly, never
admitting or settling, so concurrent callers were unbounded and the 64-frame / 64-MiB grants
bounded nothing.

`publish_record/2` now admits before any I/O and settles after:

| situation | behaviour |
|---|---|
| lane saturated | `{:error, :capacity}`, **nothing published** |
| durable PubAck | settles `:primary_publication`, credit released |
| proven poison | settles `:permanent_rejection` |
| timeout / capacity / systemic / misrouted | frame stays **outstanding** |
| re-publish of an outstanding sequence | admitted, **no second credit** |
| no pool running | **fails closed** |

The last two are the subtle ones. `PublishWindow` keeps a retryable frame outstanding on purpose
-- the publisher republishes the same bytes on the same slot -- so `:already_outstanding` is not an
error on a retry, it is the designed path; re-admitting would hand the same budget out twice.
Failing closed with no pool matters because a publish with no window is an unbounded publish,
which is the state this replaces.

The outcome passed to `settle` is an **accounting** outcome, not a disposition. This publisher
writes one record to its resolved route and does not compute audit/quarantine/security-quarantine
routing; that is task 3.5, which supplies the outcome when it lands.

## [P2] Lane connections moved to `PublisherSupervisor`, beside their pools

Under `NATS.Supervisor` they started in **every** NATS-enabled process -- core and web-ng, and
core's Helm chart enables NATS -- so three idle sockets per lane appeared wherever no pool
existed, while only the gateway started pools. One supervisor now builds both halves from one
list, so a lane cannot have a socket without a window or a window without a socket. Connections
are ordered before pools, and a test asserts that ordering.

## [P2] Credits are configurable for real

The gateway started a bare supervisor, so every deployment received the hard-coded defaults with
no channel to change them. They now come from application config (`opts` > per-lane > global >
default), and the test asserts the configured value reaches the **child spec** -- a spec builder
that ignored config would satisfy a `credits_for/2` test.

## [P2] Root composition is bound

Deleting `edge_publisher_pools_child()` left all 32 core and 37 gateway tests green. `core_children/0`
is public and asserted now.

Worth flagging how this went wrong first: my initial version asserted the **running** tree via
`Supervisor.which_children/1`. It passed locally and **failed under Bazel**, where the gateway
application is not started -- it proved where it ran, not what it composed. The assertion is now
"composed OR already running", which is the actual invariant (every core child carries an
already-started guard so core and the gateway can share a VM) and fails both ways when the entry
is deleted.

## Verification

Core 35/35, gateway 44/44, six Bazel targets green on RBE, `--warnings-as-errors` clean, format
clean, `credo --strict` clean.

Mutation-verified, each confirmed to apply before running:

| mutation | result |
|---|---|
| saturation no longer refuses | **killed** (1) |
| success never settles | **killed** (1) |
| retry refused instead of republished | **killed** (1) |
| pools dropped from the gateway tree | **killed** (1, by message) |
| configured credits ignored | **killed** (1) |
| lane connections back in the shared supervisor | **killed** (3) |
| pools ordered before connections | **killed** (1) |

## NOT closed

**Task 3.3 stays open.** `publish_record/2` still has no production caller: the
`EdgeRecordIngestService.Stream` server is the remaining half of P1, and nothing here should be
read as claiming the vertical slice. What changed is that the windows now gate any caller and the
gateway supervises the pools, so the server plugs into a bound path rather than needing to
establish one.

I did not fold the Stream server into this change. It is a bidirectional gRPC service with a
lane-open handshake, and while the ABI fields are frozen (`granted_byte_credits`,
`granted_frame_credits`), the cap policy is not -- the intended shape is
`granted = min(requested, configured lane credits)` under `1 <= granted <= requested`. That is a
substantive design worth its own review rather than an appendix to a change that already touches
supervision, transport and accounting.

Two pre-existing typing warnings remain in the gateway suite (`class == :poison` comparisons).
They are deliberate assertions that no broker-driven branch reaches `:poison`, which Elixir 1.19's
inference now also proves statically. I left them rather than weakening a real assertion to
silence a warning.


mfreeman451 (commented):
Reviewed exact head `54b53c9fa8c400437f501e9c151669a2160f1dae` against base `5a961d3c8edfa46d61a9f0fee25b5874f65da147`. The PR is now merged; these findings apply to the merged commit.

**[P1] Key reservations by the complete authenticated delivery slot, not the lane-local sequence.** `publish_record/2` extracts only `slot.sequence`, `PublishWindow` uses that integer as its sole map key, and `:already_outstanding` is accepted as a retry without comparing the remaining slot coordinates or the record identity. But the normative identity is `edge_slot = (network_scope_id, authenticated_agent_id, spool_id, sequence)`, and one lane pool is shared by every agent/spool in that class. I ran the concrete one-frame case: slot A/sequence 1 timed out, then a different agent and spool sent different bytes at sequence 1. B was published and returned `{:ok, ack}` instead of being independently admitted/refused, and B's ack released A's reservation. This can publish arbitrarily many same-numbered slots without credits and lets one record settle another's credit. Carry the full authenticated slot (plus the retained physical identity/size needed to prove a retry is the same publication) through `PublisherPool`/`PublishWindow`.

**[P1] Complete all fallible local derivation before admission.** At `jet_stream_publisher.ex:89-98`, admission happens before `headers_for/2`. A publication can have a routable contract, binary body, and positive sequence but fail the later UUID/digest/proof validation. In an exact-head test, a bad `record_sha256` returned a derivation error and performed no request, yet the one-frame pool ended at `outstanding_frames: 1, available_frames: 0`. Move header/identity derivation before `admit`, or explicitly unwind a reservation when no I/O was started.

**[P1] Re-arm a verified retry's deadline.** The `:already_outstanding` branch returns `:ok` without calling `PublisherPool.rearm/3`, despite `PublishWindow` documenting re-arm as the retry path. With an initial zero-millisecond timeout followed by a retry with a 60-second timeout, `expired/2` still returned `[1]` immediately after the retry. Once the full-slot identity check proves this is the same publication, move its deadline before issuing the retry.

**[P1] A pool restart currently releases credits while its NATS requests remain live.** `PublisherSupervisor` places connections and pools as `:one_for_one` siblings, while `PublisherPool.init/1` recreates an empty window. A caller that already admitted can remain blocked in `Connection.request/4` on the unchanged lane connection; restarting only the pool makes all capacity available again, so another caller can publish past the grant. The original caller also holds the old pool PID and can exit in the later `settle/3` call after receiving a durable PubAck. The lane's accounting and in-flight request ownership need a common restart boundary (or an equivalent mechanism that cancels/replays requests or recovers reservations) so a process crash cannot relax the bound.

**[P2] Do not start pool-only topology when NATS settings cannot be built.** `PublisherSupervisor.connection_child_specs/0` turns a settings error into `[]`, but `child_specs/1` still appends all three pools and `init/1` succeeds. That contradicts the claimed structural connection/window pairing and admits frames into windows whose named connection cannot exist. Fail startup or withhold the pools on this branch.

**[P2] Reconcile the `PublisherPool` moduledoc with this change.** Its “What is wired” section still says `ServiceRadar.NATS.Supervisor` owns lane connections and that the publisher does not admit or settle through the window. Both statements are false at this head, while “sized by its own grant” also overstates the provisional configured credits.

Independent evidence: the existing exact-head checks are green, but three focused counterexamples for cross-slot aliasing, pre-I/O leakage, and retry re-arming failed 3/3 with the concrete states above.

PR #4103, opened 2026-08-28
feat(flow-collector): shared NATS KV template store (revives #3148)

mfreeman451:
Revives #3148, re-derived onto current `staging` and rescoped.

## Why this is a re-derive, not a rebase

#3148 branched before the Forgejo -> GitHub migration re-rooted the repo. Its
branch and `staging` share **no merge base**, so there was nothing to rebase --
the two commits were cherry-picked onto current `staging` and the ~30 conflict
hunks resolved by hand.

## What was dropped, and why

The original PR flipped the collector from a `Deployment` to a per-node
`DaemonSet` and deleted its PVC on the premise that "the Rust binary writes
nothing to disk". Both are false against `staging` today:

* The publisher now runs `ensure_owned_stream` at startup -- a JetStream stream
  create/update reconciled from durable on-disk markers (`rehome_state_path`
  plus an ownership inventory) on a **ReadWriteOnce** PVC -- with no leader
  election.
* The chart pins `strategy: Recreate` precisely because two live publishers
  during a subject rehome is unsafe. A DaemonSet makes that the permanent
  steady state, and a RWO PVC cannot be shared across nodes regardless.

So the deployment shape is left exactly as `staging` has it. Making stream
ownership singleton-safe is the real prerequisite for multi-pod ingest and is
worth its own issue.

Also dropped: `k8s/demo/base/` edits (that tree no longer exists), the
publisher's `nats_client` refactor (staging's connect path already handles
mTLS/creds correctly), and `NOTES.txt` (it documented the DaemonSet migration).
The `netflow_parser` bump is moot -- staging is already on 1.0.6, the latest.

## What landed

* **`template_store.rs`** -- `NatsKvTemplateStore` over JetStream KV, bridging
  async_nats to the parser's sync `TemplateStore` trait.
* **`nats_client.rs`** -- TLS/creds-aware connect helpers for the template-store
  bootstrap.
* **Prometheus `/metrics`** -- note `metrics_addr` already existed in `Config`
  on staging but **nothing ever read it**, so port 50046 has been an exposed
  dead endpoint. This makes it real; `values.yaml` now sets `metrics_addr`.
  sFlow listeners are filtered out of the `template_store_*` and `sources`
  rows, which are structurally zero for a template-less protocol.
* **1Hz metrics ticker** with retired-source accounting, replacing the original
  per-datagram O(sources) scan under the parser lock (qodo raised this on the
  original PR).

## Two fixes on top

1. **Bounded KV ops** -- the trait is synchronous and the parser calls get/put
   *while holding the parser mutex*, so each round-trip stalls all parsing for
   that listener. async-nats' default request timeout is 5s, far too long to
   hold the parse lock. Capped at 500ms; an overrun is reported as a `Backend`
   error, which the parser already treats as a cache miss. A NATS outage now
   degrades to in-process-only caching instead of stalling ingest.

2. **NATS KV permissions** -- found only by deploying. The flow-collector user
   is allowed `$JS.API.>`, which covers `create_or_update_key_value`, but KV
   get/put/delete ride on `$KV.<bucket>.>` directly and those were in neither
   the publish nor subscribe allow list. The feature came up looking healthy
   -- bucket present, ticker running -- and silently stored nothing.

## Verified on farm01

Deployed to the farm01 cluster (`serviceradar` ns) on this build.

Before: `/metrics` refused connections (the dead port), `KV_flow_templates`
did not exist.

After the permissions fix, 12 -> 28 template entries written, and a pod
restart produced:

```
Template restored from secondary store - ID: Some(259), Protocol: Ipfix
Template restored from secondary store - ID: Some(261), Protocol: Ipfix
Template restored from secondary store - ID: Some(262), Protocol: Ipfix
```

Those are the **same** template IDs that logged `data lost` on the previous
cold start. `flow_collector_template_store_restored_total` read exactly 3,
matching the log lines, with 0 codec and 0 backend errors and 0 dropped flows.

`make test`: 200 tests pass. `cargo clippy --all-targets -D warnings`: clean.

Signed-off-by: Michael Freeman <mfreeman451@gmail.com>


mfreeman451:
Follow-up work split out into #4107: the collector cannot safely run more than one pod (per-pod `ensure_owned_stream` with no leader election, durable markers on a RWO PVC), and separately `netflow_parser`'s `DEFAULT_MAX_SOURCES` is 10,000 with the collector never calling `with_max_sources()` — a hard ceiling well under the ~25k-exporter SaaS target.

The shared template store landed here is a prerequisite for that work, not a substitute for it: it makes LRU source eviction survivable instead of data-losing.

PR #4110, opened 2026-08-28
fix(edge): key reservations on the PUBLICATION; provisional admissions and attempt phases

mfreeman451:
Fix-forward on the merged `54b53c9` and its successors, opened against the current integration head.

Reservations are keyed on the **publication** — the authenticated slot AND the record identity — and an admission passes through a **provisional** phase before it can be acted on. Both came out of review; the history below is the short version of why each shape is what it is.

## What the keying has to satisfy, in both directions

* Keyed on the lane **sequence**, different agents alias: one lane pool serves every agent and spool in its class, so agent B at sequence 1 looked like agent A retrying, published on A's credits, and its ack released A's reservation.
* Keyed on the **slot**, a second record on that slot gets refused — but the spec REQUIRES it to be published: `Nats-Msg-Id` binds `record_sha256`, so JetStream does not deduplicate it away and *"the frame SHALL reach EventWriter"*, which rejects it as a transport-integrity violation. Refusing it in the gateway moves that adjudication upstream and destroys the evidence.

Keying on the publication satisfies both. A retry is the same key and costs nothing extra; a different record is a different key, admitted or refused on its own credits like any other frame.

## Reservations, attempts, and phases

A **reservation** holds credits. An **attempt** is what is in flight against them. Conflating them let concurrent retries share one charge.

| | outcome |
|---|---|
| admit a new publication | reservation + provisional attempt |
| admit while an attempt is live | `:attempt_in_flight` |
| admit after the attempt ended | retry: re-arms, no new credits, NEW token |
| confirm | activates the attempt |
| `attempt_failed/2` | keeps credits, ends the attempt |
| `settle/3` | releases |
| `abandon/2` / `revoke_pending/2` | undo an admission its caller never received |

Attempt tokens come from `System.unique_integer/1`, so they never repeat across a pool incarnation — a per-window counter made a stale reservation compare equal to a fresh one after a restart.

**Provisional attempts are inert.** They hold their slot but are invisible to `expired/2` and refused by `settle/3`, `rearm/3` and `attempt_failed/2`. A provisional token reported as expired was enough to end the attempt, admit a retry, and put two requests on the wire under one charge before the first caller had received its reservation.

## Admission handoff

`GenServer.call/3` exits the caller on timeout, but OTP does not cancel the queued message — so the pool could admit a frame for a request that was never made. Each admission carries a reference the caller revokes on its own timeout; FIFO ordering means the revocation is handled after the admission it cancels. Revocation is bound to the exact token and to whether the admission **created** or **re-armed** the reservation: a re-armed one is restored, not released, because the record is still unresolved.

The pool monitors a caller only until the handoff is confirmed. Death before that is provably pre-publication and revocable; death after it is conservatively charged, because a request may be on the socket. **Caller death authorises nothing on its own.**

## Verification

Core edge+nats **1172/1172**, gateway **213/213**, six Bazel targets green on RBE, warnings-as-errors clean, format clean, credo clean. Every fix in this PR is mutation-verified with each mutation confirmed to apply before running.

## NOT closed

**Task 3.3 remains open**: `publish_record/2` still has no production caller, so all of this is latent until the `EdgeRecordIngestService.Stream` server is wired.

**The hard window is NOT satisfied.** Two races remain, and both are TASK 3.3's own obligation — not 3.4's (exact-byte and retained-memory binding) and not 3.5's (outcome-specific PubAck validation and prefix advancement):

1. **Restart over-admission.** Supervisor restarts are eventual, so between a pool crashing and its sibling terminating, a publisher holding the old connection can complete a request while the replacement pool starts with its full grant.
2. **Post-handoff.** The window bounds ADMISSIONS, not the interval between a caller receiving its reservation and reaching the socket. A caller can receive attempt 1, be descheduled past its deadline, have a sweep end attempt 1 and admit attempt 2, then resume and publish under attempt 1. Two publications, one charge — ordinary scheduling, not a crash.

What IS closed is the reporting: a publish whose settlement cannot be recorded returns a **retryable** class rather than being reported durable. What that class *obliges* is a caller's business — this layer holds no source-sequence state and advances or withholds nothing. Nothing here republishes either (there is no production caller), and `Nats-Msg-Id` deduplication is not a general answer, being scoped to one stream and one duplicate window; the proposal names database idempotency as the backstop.

**Correlation alone will not fence the post-handoff race.** The absence of a PubAck cannot distinguish "never sent" from "in flight", "delayed", or "acked but the ack was lost". 3.3's publisher pipeline needs fencing on the request itself — owner, start, termination — before a retry may be admitted.


mfreeman451:
Pushed `9980122116`. All three P1s and all three P2s addressed. The first finding showed my previous fix was itself wrong, so the keying is rebuilt rather than patched.

## [P1] `:slot_conflict` was a spec violation, and is gone

It blocked a case the spec **requires** to publish:

> **WHEN** a frame reuses the same slot with a different `record_sha256` R2 ... **THEN** because `Nats-Msg-Id` binds `record_sha256`, JetStream SHALL NOT deduplicate it away and the frame SHALL reach EventWriter
> **AND** EventWriter SHALL reject it as a transport-integrity violation

Refusing in the gateway moved EventWriter's adjudication upstream and destroyed the evidence. Reservations now key on the **publication** -- the authenticated slot **and** the record identity:

| | outcome |
|---|---|
| same publication | retry: no new credits, deadline re-armed, **same token** |
| different record, same slot | a different publication, admitted on its **own** credits |
| different agent/spool | independent reservation |

Both directions fall out of one key, and no `:slot_conflict` remains. A gateway test asserts the second record publishes with a distinct `Nats-Msg-Id`.

## [P1] The attempt no longer straddles a lane restart

The connection is resolved to a **pid** before admission, and the request uses that pid rather than re-resolving the name. A captured pid is dead after a restart, so the publish fails instead of going out unaccounted through the new connection. Settlement is guarded too -- a pool that dies mid-attempt is reported, not raised at the caller. Tested by killing the pool *during* the request.

## [P1] Settlement ABA

Each reservation carries an **epoch token**. Concurrent retries share one reservation and one token; once settled, a late ack naming that epoch cannot release a later reservation that reused the key.

## [P2] All three

`expired/1` takes no clock -- the pool reads its own, so `expired(pool, nil)` is unrepresentable rather than a crash that took the lane down with it. The expired-then-rearm ABA closes on the same token. And the deadline is stamped by the **pool** when it admits, so time queued for a contended pool is no longer deducted from the PubAck interval.

## Verification

Core edge+nats **1152/1152**, gateway **210/210** across two seeds, six Bazel targets green on RBE, warnings-as-errors clean, format clean, credo clean.

Mutations, each confirmed to apply first: key drops record identity (2 failures), settle ignores token (2), rearm ignores token (2), pool stamps `now` (1), publisher re-resolves the name (1), settlement unguarded (1).

Worth flagging: the conservation tests could no longer assert `settled === start`, because `next_token` is monotonic by design -- a window that conserves credits perfectly still differs. They now compare the accounting fields, with the reason recorded, rather than being quietly weakened.

## Not addressed, per your note

The gateway/NATS `already_started` boot defect is left alone as pre-existing and separate.

**Task 3.3 remains open** -- `publish_record/2` still has no production caller.

mfreeman451:
Pushed `06a4b339a7`. All three P1 blockers plus the additional P1 and all three P2s.

The root of the first finding was a conflation: a **reservation** holds credits, an **attempt** is what is in flight against them. Treating them as one thing is what let two requests share one charge.

## [P1] Concurrent identical retries

Every same-key admission returned the same reservation and the publisher immediately started another request. A second admit while an attempt is in flight is now refused `:attempt_in_flight`. The publisher reports every non-terminal outcome with `attempt_failed/2`, which **keeps the credits** — the record is still owed a republish — and ends the attempt so the retry is a legal re-admission with a **new** token.

## [P1] Timed-out admission charging the pool

This is the one I'd have missed reasoning from crashes: `GenServer.call/3` exits the caller on timeout and OTP does not cancel the queued message. The pool now monitors admitting callers and abandons what they held on `:DOWN`. `abandon/2` is a separate named entry point, not a settlement — nothing was published and no disposition applies. Tested by admitting from a process that is then killed, polling until the credit returns, and confirming another caller's reservation is untouched.

## [P1] Epochs repeating across restarts

`next_token` reset to 1 with the window. Tokens now come from `System.unique_integer/1` and are minted **per attempt**, so a stale handle can neither settle a fresh reservation nor end a superseded attempt.

## [P1] Lane restart exiting `publish_record/2`

The pid was captured but admission was unguarded. Every pool call is guarded now, and a dead pool at admission returns `{:error, :systemic}` with nothing published.

## [P2] All three

The same-slot/different-record test keeps A **outstanding** across B with two credits — you were right that settling A first would pass just as well with record identity dropped from the key. Public contracts reconciled (`key/5`, no `:slot_conflict`, `outstanding?/2` retyped, settle docs now stating what the reservation *does* carry and what it still does not — the destination, owed by 3.5).

On the lane supervisor: you were right that I overstated it. The supervised child is Gnat's **reconnect manager**, not the transport, so an ordinary reconnect does not exit it and `:one_for_all` does not fire. Reservations surviving a reconnect is correct — a reconnect does not make the records go away, and the failed requests are reported through `attempt_failed/2`. I documented that as deliberate and renamed the test for what it actually kills, rather than claiming a transport-generation guarantee the code does not make.

## Verification

Core edge+nats **1159/1159**, gateway **47/47**, six Bazel targets green on RBE, warnings-as-errors clean, format clean, credo clean.

Mutations, each confirmed to apply first: concurrent retries allowed (2 failures), tokens repeat (7), dead caller strands its credits (2), unguarded admission (1), attempt never ends (2).

**Task 3.3 remains open** — `publish_record/2` still has no production caller.

mfreeman451:
Pushed `f05bd35acb`. Both P1s and both P2s.

You were right that the monitor design was wrong in **principle**, not detail — so it's gone entirely rather than patched.

## [P1] The queued-call timeout was still unfixed

The diagnosis I missed: `pool_call/1` catches the exit, so the caller **survives** — and monitoring a live caller recovers nothing. My "fix" only worked in the test because the test killed the caller by hand.

Each admission now carries a reference the caller can **revoke**. On its own call timeout, `PublisherPool.admit/4` casts `{:cancel_admission, ref}` and returns `:pool_timeout`. Message order between the caller and the pool is FIFO, so the revocation is handled *after* the admission it cancels — whether that admission had already been processed or was still queued.

What authorises the release is the thing that was missing before: the caller never **received** the reservation, so it provably never published.

The test now reproduces your case rather than a proxy for it — `:sys.suspend` the pool, let the call time out, resume, and assert the credit comes back **with the caller still alive**.

## [P1] Caller death authorises nothing

Removed. A caller can die after Gnat has written to the socket, or exit normally after `attempt_failed/2` deliberately kept the credit; releasing then permits a second publish while the first is broker-ambiguous. Keeping credits charged is the conservative direction, and a lane restart is what clears them. Two tests pin this: killing a caller releases nothing, and a stray `:DOWN` is inert.

That also disposes of the ignored-monitor-reference bug and the P2 monitor leak — there are no monitors left to leak.

## [P2] Contracts

Stored attempt token typed `pos_integer() | nil`; `admit/5` corrected to `admit/4`; `expired/2` documented as reporting **reservations**, and as skipping a reservation between attempts; the surface guard now says settle/3 **and** abandon/2 release, with abandon/2's narrow authority spelled out — it takes a key rather than a reservation precisely because no live attempt holds a token for it.

## Verification

Core edge+nats **1161/1161**, gateway **213/213**, six Bazel targets green on RBE, warnings-as-errors clean, format clean, credo clean.

Mutations, each confirmed to apply first: no revocation (2 failures), revocation ignores the reference (1), caller death releases again (1).

One deliberate addition: the pool's call timeout is now a named, configurable value rather than the invisible `GenServer` default — the revocation protocol is built around it, and it needs to be tunable to exercise that path without a five-second test.

**Task 3.3 remains open** — no production `publish_record/2` caller, so these paths stay latent until it is wired.

mfreeman451:
Pushed `bd72ab24e3`. Both P1s and all three P2s.

## [P1] Revoking a retry released an unresolved reservation

This was a capacity bug I introduced with the revocation protocol. A retry adds no credits — it re-arms a reservation still owed a republish — so undoing it must restore the no-attempt state, not abandon it.

Revocation now records whether the admission **created** or **re-armed**, and is bound to the exact attempt token: `:created` abandons, `:rearmed` calls `attempt_failed/2`, and a superseded token undoes nothing at all.

## [P1] Death before handoff left a phantom attempt

Your framing is what made this tractable: publication cannot happen before `admit/4` returns, so a caller that dies while its admission is still queued provably never published — and *that*, not liveness in general, is what authorises release. No PubAck correlation needed.

The pool now monitors a caller **only until the handoff is confirmed**: the client casts `{:confirm_admission, ref}` once it holds the reservation, and the monitor is dropped. Death after confirmation stays conservatively charged, because by then a request may be on the socket. Erlang's signal ordering makes the confirm-then-die case safe — the cast is already queued ahead of the `:DOWN`.

## [P2] All three

Bookkeeping is now bounded by in-flight admit calls rather than retries — an entry leaves on confirmation, revocation, or pre-handoff death. A missing pool is `:systemic` rather than `{:derivation, _}`, since a restart gap is exactly that shape and `:derivation` means a bad route/identity/grant.

`abandon/2` takes a **reservation** now, so it can enforce its own contract — you were right that a key-only function cannot express "unreceived admission only". Its doc no longer claims owner death authorises it.

The timeout tests moved to `PublisherPoolHandoffTest` with `async: false`; mutating global config from an async module was wrong. And the deadline-after-queue claim is now bound by a test that fails if the ack window is charged for queue time — previously it would have passed either way, as you noted.

## Verification

Core edge+nats **1164/1164**, gateway **213/213**, six Bazel targets green on RBE, warnings-as-errors clean, format clean, credo clean.

Mutations, each confirmed to apply first: revoking a retry abandons (1 failure), no pre-handoff monitor (1), confirmation never drops pending (2), queue time deducted from the ack window (1), missing pool as derivation (1).

One test-design note: I dropped a non-vacuity probe that tried to observe a pending entry mid-flight. Confirmation is a cast from the same process, so the entry is already gone by the time a test can look — the probe was racing itself. What actually proves the map is live is that revocation works at all, since revocation can only find an admission through it.

**Task 3.3 remains open** — no production `publish_record/2` caller.

Separately: I'm looking into the `integration_tests_serial_1` failure you flagged and will report what it is rather than leave it as "unrelated".

mfreeman451:
Pushed `fd86df3203`. Both P1s addressed — one closed, one closed *in part* with the residue stated.

## [P1] Provisional admissions

You were right that the handoff was only half-built: the token existed before its caller did. An attempt now has a phase. Admission creates `{:pending, token}`; confirmation activates it. Provisional attempts hold their slot — so no retry can occupy it — but are invisible to `expired/2` and refused by `settle/3`, `rearm/3` and `attempt_failed/2`.

`revoke_pending/2` is the retry half of revocation, named for its authority: only an incomplete handoff may use it, and it restores rather than releases.

## [P1] The restart fence — closed in part, and I want to be exact about which part

You're right that `:one_for_all` is not atomic, and my moduledoc claimed otherwise. I could not close the over-admission itself: between a pool crashing and its sibling terminating, a publisher holding the old connection completes its request, and the replacement pool has no way to know that. A generation fence would have to be honoured by something on the publish path, and JetStream does not enforce a local credit bound.

What I did close is the **reporting**: a publish whose settlement cannot be recorded — lane restarted, or reservation superseded — is no longer returned as durable. Progress is withheld and the record republished, with `Nats-Msg-Id` deduplicating the copy. Reporting `{:ok, ack}` there was reporting a fact nothing could account for.

The residual is now stated in `LaneSupervisor`'s moduledoc and in the PR body, scoped to the PubAck correlation in 3.4/3.5. I'd rather leave it named than let the supervisor imply a guarantee it doesn't provide.

## Smaller findings

All four. The retry-bookkeeping test now asserts no **monitors** leak, not just an empty map. The concurrency tests wait until the call has actually reached the suspended pool rather than sleeping a fixed interval — you were right that a fixed sleep proves nothing about the interleaving. The deadline test measures the **recorded deadline** against a 30s window instead of racing a 200ms one, so slow scheduling cannot fail correct code. The public-surface message no longer says `abandon/2` takes a key, and the PR body is rewritten — it still described `:slot_conflict` and `admit/5`, neither of which exists here.

## Verification

Core edge+nats **1168/1168**, gateway **213/213**, six Bazel targets green on RBE.

Mutations, each confirmed to apply first: expired exposes provisional attempts (1 failure), settle accepts a provisional attempt (1), confirmation does not activate (2), an unaccountable publish reported durable (1).

Thanks for running down `integration_tests_serial_1` — I had it queued to investigate and would have spent a while on it.

mfreeman451:
Pushed `a55a8fbc38`. P1-1 closed; P1-2 scoped rather than claimed.

## [P1] `abandon/2` released active attempts

You're right, and the test was worse than the code: it **activated** the attempt and then asserted abandonment succeeded, so it enshrined exactly the bug. In a one-credit window that is activate A, abandon A, admit B — while A may already be publishing.

Abandonment is now restricted to `{:pending, token}`, the only state its documented authority covers, and the test is inverted: an active attempt must be refused. Mutating it back to `{_phase, token}` fails.

## [P1] The post-handoff race — documented, not claimed

I did not close this and I'm not going to pretend otherwise. The window bounds **admissions**; it cannot see the interval between a caller receiving its reservation and reaching the socket. Your interleaving — receive token 1, deschedule, deadline passes, a sweep ends token 1 and admits token 2, original caller resumes and publishes — is ordinary scheduling, and every gate here is on the reservation.

`PublishWindow`'s moduledoc now carries a "WHAT THIS DOES NOT YET BOUND" section stating the interleaving, why no gate here can see it, that closing it needs the retry fenced against the original attempt's owner, and that **the hard-window requirement is not satisfied until then**.

## Smaller corrections

All five. `LaneSupervisor` no longer says in-flight requests "are dropped with" the connection — that contradicted its own caveat. The reporting comment no longer claims this function republishes (it returns a retryable error; there is no caller) nor that `Nats-Msg-Id` dedup is a general answer — it's scoped to one stream and one duplicate window, and the proposal names database idempotency as the backstop.

The premature-activation regression goes **through the pool**, calling the server directly, since `admit/4` confirms for you and a pure-window test never sees it — only the pool asks the window to activate. And the deadline test now captures the instant before resuming and asserts `deadline >= resumed_at + ack_window`; mutating the stamp by **one millisecond** now fails it, where the old 200ms budget would not have.

PR title corrected to say **publication**, not "complete slot".

## Verification

Core edge+nats **1170/1170**, gateway **213/213**, six Bazel targets green on RBE.

Mutations, each confirmed to apply first: abandon accepts an active attempt (1 failure), the pool activates at admission (4), deadline stamped 1ms early (1).

## Standing scope

Task 3.3 stays unchecked and the hard window is **not** complete. Two named prerequisites remain for 3.4/3.5: the restart over-admission window, and the post-handoff race above. Neither should be wired into the vertical slice as-is.

mfreeman451:
Pushed `9f4e039329` — documentation and claims only, no behaviour change.

## Ownership was wrong, and I'd repeated it in four places

Both races violate **task 3.3's** hard window. 3.4 owns exact-byte and retained-memory binding; 3.5 owns outcome-specific PubAck validation and prefix advancement. Correlation may assist recovery but does not move the obligation. Every "owed by 3.4/3.5" is corrected in `PublishWindow`, `PublisherPool`, `LaneSupervisor` and the publisher.

## Correlation alone would not fence it either

Also corrected, and this is the part I had wrong on the merits rather than just the attribution: the absence of a PubAck cannot distinguish "never sent" from "in flight", "delayed", or "acked but the ack was lost". The moduledoc now says 3.3's publisher pipeline needs fencing on the **request** — owner, start, termination — before a retry may be admitted, instead of pointing at correlation as though it were sufficient.

## Contradictory claims removed

`PublishWindow` said its accounting "makes that safe" and bounds outstanding work "at any moment" — it bounds **admissions**, which its own new section shows is not the same thing. `PublisherPool` called it "a real bound across concurrent callers". `LaneSupervisor` said records are withheld, republished and offered again, when this layer only returns a retryable error.

## The test claim, narrowed

You're right, and this one matters to me more than the doc fixes: my `-1ms` mutant was killed by **timing, not determinism**. If the pool begins handling a millisecond after `resumed_at` — which is usual — that mutant passes. Reporting it as "confirming the bound is exact rather than approximate" was an overclaim about my own verification.

The test comment now states what it does establish (it catches a caller-stamped deadline losing the injected ~300ms of queue time) and what it does not (exact arithmetic, which would need the clock injected into the pool). I chose to narrow the claim rather than add another production seam, but say the word if you'd rather have the injection and the exact assertion.

## PR body

Rewritten: the post-handoff blocker is now listed alongside the restart one, both attributed to 3.3, with the correlation caveat. Test count corrected to 1170.

## Verification

Core edge+nats **1170/1170**, gateway **213/213**, six Bazel targets green on RBE, warnings-as-errors clean, format clean, credo clean.

**Task 3.3 remains open with two named prerequisites of its own.** The vertical slice stays blocked.

mfreeman451:
Pushed `8423f60694`, rebased onto the current integration head.

## The blockers are now normative

This was the right thing to insist on — they lived only in moduledocs and this thread, which is exactly how a future pass checks 3.3 having proved nothing more than bounded reservations.

**`specs/ingestion-routing`**, under "Backpressure and fairness are bounded at every hop", gains two scenarios:

- *A publisher lane restarts with requests in flight* — the replacement accounting SHALL NOT reopen capacity an in-flight request still occupies, and eventual supervisor restart of a sibling SHALL NOT be treated as satisfying it, because restarts are ordered but not instantaneous.
- *A retry is offered while the previous attempt may still publish* — no retry on that reservation until the previous attempt is fenced by its **request** (owner, start, termination). A passed deadline or absent PubAck SHALL NOT authorize it, since neither distinguishes "never sent" from "in flight", "delayed", or "acknowledged with the acknowledgement lost".

**`tasks.md` 3.3** gains explicit closure criteria naming both, stating plainly that bounded reservations are not the hard window, and recording that this is 3.3's own obligation — not 3.4's exact-byte/retained-memory binding, not 3.5's outcome-specific PubAck validation. `openspec validate --strict` passes.

## The two comment corrections

`PublisherPool` assigned asynchronous pipelining to 3.4/3.5 — pipelining, the hard window, and recording out-of-order PubAcks are all 3.3.

`LaneSupervisor` and `JetStreamPublisher` said the source sequence "stays unresolved"/"stays withheld". This layer returns a retryable error and holds no such state; withholding progress is the future caller's obligation. I reworded the two refusal-classification comments the same way, since they carried the identical implication.

## Verification

Rebased onto `usp-01-proposal`; core edge+nats **1172/1172**, gateway **213/213**, six Bazel targets green on RBE, warnings-as-errors clean, format clean.

## Standing scope

Task 3.3 stays unchecked with two named prerequisites **of its own**, now enforceable from the spec rather than from memory. The vertical slice stays blocked.

Thanks for holding the line on the assignment — "owed by 3.4/3.5" would have quietly turned 3.3's unfinished work into someone else's problem, and I'd repeated it in four places without noticing.

mfreeman451:
This review is against `9f4e039329`, which is one revision behind — all three items were addressed in **`8423f60694`**, pushed shortly before it arrived. Verified by content at the remote head rather than asserted:

| item | state at `8423f60694` |
|---|---|
| 1. `PublisherPool` pipelining ownership | reworded: *"pipelining that would make out-of-order settlement possible is also TASK 3.3's"*, with 3.4 (exact-byte/retained-memory) and 3.5 (outcome-specific PubAck validation) named for contrast |
| 2. "stays unresolved" / "stays withheld" | **0 occurrences** in `LaneSupervisor` and `JetStreamPublisher` — including the two refusal-classification comments that carried the same implication |
| 3. OpenSpec | both scenarios present at `spec.md:373` and `:387`; `tasks.md` 3.3 carries `CLOSURE CRITERIA`; `openspec validate --strict` passes |

The branch is also no longer behind: rebased onto `usp-01-proposal`, which is now an ancestor (0 commits behind).

The two scenarios landed under "Backpressure and fairness are bounded at every hop" as:

- **A publisher lane restarts with requests in flight** — the replacement accounting SHALL NOT reopen capacity an in-flight request still occupies; eventual supervisor restart of a sibling SHALL NOT be treated as satisfying it, because restarts are ordered but not instantaneous.
- **A retry is offered while the previous attempt may still publish** — no retry until the previous attempt is fenced by its request (owner, start, termination); a passed deadline or absent PubAck SHALL NOT authorize it, since neither distinguishes "never sent" from "in flight", "delayed", or "acknowledged with the acknowledgement lost".

Post-rebase verification: core edge+nats **1172/1172**, gateway **213/213**, six Bazel targets green on RBE.

Nothing further from me on these three — please re-review at `8423f60694`.

mfreeman451:
Pushed `9d6dbab5c4`. Text only.

**My verification was the error, not just the text.** I searched for the exact phrases, got zero hits, and reported the claim gone. A phrase-match is not a claim-check, and every semantic variant you listed survived it. This pass searched for the *claim* — anything asserting this layer resolves, withholds, republishes or deduplicates — and corrected all of it:

`jet_stream_publisher.ex` at :33, :192, :207, :451, :455, :467, plus two more the same shape at :480 (`"it withholds rather than DLQs"`) and :509 (`"always withheld"`), which your line list didn't include but carry the identical implication. The test comment at :478 and a test name (`"is unresolved"` → `"is retryable"`). PR body: count corrected to **1172**, and the "stays unresolved" sentence rewritten to say the retryable class *obliges a caller*, while this layer holds no such state.

## One thing this run surfaced that you should see

24 core edge tests failed — credential and remote-access paper-trail tables, `column "version_action_inputs" ... does not exist`. Not from this diff: it touches no migrations and no credential code, and the same files passed **1172/1172** against an earlier clone. The cause is that **`sr_core_template` has moved ahead of the branch again** — template max `20260830220000` against `20260825031000` on disk — so a later migration reshaped `platform.network_credential_secret_versions`.

This is the same incident that opened this workstream. **PR #4061, the drift guard built for exactly this, is CLOSED unmerged** — so the second occurrence again presented as two dozen opaque `MatchError`s instead of one line naming the extra applied migrations. Worth reviving independently of this PR; it will keep costing an investigation each time.

Verification for this diff: publisher/window/pool/lane suites **99/99**, gateway **213/213**.

Issue #4107, opened 2026-08-28
flow-collector cannot scale past a single pod: stream-ownership singleton + hard 10k exporter ceiling


mfreeman451:
Splitting this out of #4103 (which revived and rescoped #3148).

## Why this matters now

The SaaS target is on the order of **25,000 routers** exporting NetFlow/IPFIX
into one deployment. The flow collector today runs as a **single-replica
Deployment and cannot safely run more than one pod**. Two independent things
break before we get near that number, and one of them is a hard ceiling rather
than a performance curve.

Current utilisation is not the argument against urgency here -- farm01 sits at
8m CPU and demo at 4m, both against a 500m limit. The problem is structural,
and the lead time on fixing it is long.

## Blocker 1: hard 10,000-exporter ceiling (verified)

`netflow_parser` caps tracked sources per parser:

```rust
// scoped_parser.rs:24
/// Prevents unbounded memory growth from spoofed source addresses.
pub const DEFAULT_MAX_SOURCES: usize = 10_000;
```

It is an LRU -- sources past the cap are **evicted**, not rejected. The crate
exposes `AutoScopedParser::with_max_sources()` to raise it, and
`rust/flow-collector/` **never calls it**. Grep for `max_sources` under
`rust/flow-collector/src/` returns nothing.

Note the existing `max_templates` config knob is *not* this: it maps to
`with_cache_size` (`netflow/mod.rs:117`), which sizes the per-source *template*
cache. There is no knob for source count at all today.

So at 25k exporters against one pod, 10k are tracked and the remaining 15k
churn through continuous eviction and re-admission. Every re-admission is a
cold parser: template miss, then a NATS KV round-trip.

The shared template store from #4103 makes that survivable rather than
data-losing, which is exactly why it was worth landing first -- but it is not
a substitute for having enough parsers.

## Blocker 2: the collector is a control-plane singleton

`Publisher::ensure_owned_stream` (`rust/flow-collector/src/publisher.rs`) runs
**per pod at startup**. It creates/updates the `flows` JetStream stream and
reconciles the owned subject list from durable on-disk markers:

* `rehome_state_path` -> `/var/lib/serviceradar/flow-collector-rehome.json`
* an ownership inventory -> `/var/lib/serviceradar/flow-collector-ownership.json`
* `ready_state_path` -> `/var/lib/serviceradar/flow-collector.ready` (the
  readinessProbe gates on this file existing)

[...]

PR #4132, opened 2026-08-29
chore: release v1.4.47

mfreeman451:
Release cut for **v1.4.47**. `VERSION`, `helm/serviceradar/Chart.yaml`, the demo
ArgoCD source, and the `CHANGELOG` entry are updated by `scripts/cut-release.sh`.

## Headline

Unblocks running more than one flow collector (#4107), and adds an MCP server so
agents can query SRQL through existing SSO.

The flow-collector work is the substantive part. Three things blocked multi-pod:

* Templates were per-pod, so any pod change dropped flows until each exporter
  re-announced (~60s, RFC 3954). They now write through to a NATS JetStream KV
  bucket any pod can read.
* The publisher did a JetStream create/update reconciled from durable markers on
  a ReadWriteOnce PVC, with no leader election. That cutover moves into a
  `pre-install,pre-upgrade` hook Job, so pods derive their subject list from
  config alone and concurrent ensures converge rather than race.
* `netflow_parser` caps tracked exporters at 10,000 and evicts (LRU) past it,
  with no way to raise it. Now `listeners[].max_sources`.

## This release unblocks farm01

Chart 1.4.47 is the **first published chart carrying the NATS
`$KV.flow_templates.>` grant** (from #4103). Verified against the registry:

```
staging chart source     -> 5 occurrences of flowTemplateBucket
published chart 1.4.46   -> 0
```

Without that grant the flow collector creates its KV bucket fine and then has
every write denied, so the template store looks healthy and stores nothing.
farm01's gitops values (carverauto/gitops#56) already carry the
`template_store` config and are waiting on this chart.

## Gates

* `make test` on this exact tree: **200/200 pass, exit 0**.
* No duplicate migration versions.
* `MIX_ENV=prod mix release --overwrite` for `serviceradar_core`: assembled, no
  `duplicated modules`.
* v1.4.46 confirmed genuinely released (GitHub release published 2026-08-27,
  semver images `v1.4.46` present in Harbor) - not a tagged stub.

## Merge order matters

The annotated tag `v1.4.47` exists **locally only** and was deliberately not
pushed. `release.yml`'s "Enforce release source" guard requires the tagged commit
to be reachable from `origin/staging`, so the tag must go last.

**Merge this with a merge commit** (`--merge`), not squash or rebase - squashing
orphans the tagged commit and the guard stays red. Then push the tag.

Signed-off-by: Michael Freeman <mfreeman451@gmail.com>


mfreeman451 (commented):
lgtm

PR #4133, opened 2026-08-29
fix(edge): bound gateway resolved-prefix retention

mfreeman451:
Preserves the unmerged resolved-prefix retention follow-up found in sr-tracker-wip.\n\nThe patch adds reported-through evidence release, an explicit evidence-released refusal, bounded-retention coverage, release-bound checks, and resolving-kind supersession coverage.\n\nVerification: the focused Elixir file passes 25 tests. This is intentionally draft: valid replay after evidence release still lacks a production recovery path, and no reclaim watermark is implemented here.


mfreeman451:
The safe subset has been ported onto current `usp-01-proposal` in #4141. Keep this PR as the exact preservation snapshot and do not merge it unchanged: #4141 intentionally excludes `reported_through/2` and `:evidence_released`, while adding the OpenSpec lifecycle requirements that must be satisfied before evidence eviction is integrated with the real Stream/publisher owner.

PR #4134, opened 2026-08-29
feat(credentials): deliver NetBox and Proxmox credentials from the database

mfreeman451:
Two integration credentials could not be delivered from the database at all. Refs #4209.

## NetBox

`netbox/plugin.yaml` declared no `integrations.credential_profiles`, so NetBox could not appear in the credential UI and its API token was designed to sit **in plaintext** in `plugin_assignments.params` — a direct violation of the DB-only credential policy. The live demo assignment has `params = {}`, which is why it reports `has no sources configured`.

- Adds a `netbox` credential profile with `api_token` auth and an `inventory_sync` purpose, mirroring `unifi-protect` (the AWX inventory-sync `controllers` array turned out to come from a bespoke Elixir reconciler, not a manifest, so it was not a usable template).
- `target_cardinality: single` — a NetBox instance does not map to a target at all; `base_url` comes from the rule. Chunking a whole-instance sync **repeats** it rather than dividing it, so one rule on one agent was producing one complete `snapshot_complete=true` snapshot *per chunk*.
- The default `target_query` was `in:devices metadata.netbox_candidate:true` and **nothing in the tree writes that key** (unlike `proxmox_candidate`, which has a real writer), so it matched zero devices on every deployment. Replaced with a default that resolves.
- The failure message now names the missing piece instead of reporting "no sources" for every cause.

## CA trust material

Proxmox inventory enrichment mandates `tls_policy: verify`, `canonical_origin/3` forces an **IP-literal** origin, and `validatePluginHostAuthorityBinding` rejects a proxmox binding carrying `insecure_skip_verify` outright. Pinning the cluster CA is the only way to reach such a node — so the requirement was unsatisfiable before this.

Measured against demo pve02/pve03:

```
SANs:   IP Address:10.0.0.3, DNS:pve02, DNS:pve02.localdomain
pinned: Verify return code: 0 (ok)        # cluster CA as sole anchor, -verify_ip 10.0.0.3
none:   unable to get local issuer cert   # negative control -- today's demo
```

PVE already issues node certs with an IP SAN, so **no certificate re-issuance is needed**.

- `ca_bundle_pem` / `server_cert_fingerprint` on the rule, mutually exclusive, validated at the changeset **and** by a DB check constraint. Validated at save time rather than first use — a rule that only fails inside a plugin run hours later is the reporting failure this exists to remove.
- Stored in the clear: a CA certificate and a fingerprint are trust anchors, not authenticators, and an operator must be able to read back what a rule trusts. They never go through the broker.
- The bundle rides the **host-authority binding** beside `insecureSkipVerify` — deliberately not `httpRequestPayload`, which is guest-controlled; a plugin must not be able to nominate its own trust anchor. Added to `@secret_keys` so it is stripped before the guest sees params.
- Pinned roots **replace** the system pool rather than extend it. A rule pinning a private CA is asking for that anchor; keeping the public roots would still accept any publicly-trusted certificate for the same origin.
- Form controls for both fields.

## Verification

| Check | Result |
|---|---|
| `TrustMaterial` unit tests | 11/11, against the real demo cluster CA |
| Mutation: `validate_bundle` always throws | 3 failures — load-bearing |
| CA pinning e2e (Go) | system roots **fail** → pinned CA **succeeds** → unrelated CA **fails** |
| `go test ./go/pkg/agent/` | ok, no regressions |
| core `plugins/` + `credentials/` | 445 tests, 0 failures |

The Go test builds a private CA and a leaf with an **IP SAN** — the shape PVE issues — so it reproduces the demo hermetically. The negative control matters: without it, a test that merely succeeds with a CA pinned would also pass if pinning were silently ignored.

Not run: `make test` (full Bazel sweep).

## Known gaps

- A **gateway- or partition-scoped** NetBox rule still fans out one full snapshot per agent. `target_cardinality: single` fixes chunking within one `(rule, agent)`, not across agents.
- `api_token_secret_ref` is load-bearing but unvalidated — resolution works only because `agent_config_generator.ex` synthesizes a secretRef for four hardcoded names. Renaming it in a manifest passes validation and fails silently at delivery. Commented, not fixed.

## Merge order

Merge **after** #4136 and rebase onto it — both touch `network_credential_rules_live.ex` in different regions.


mfreeman451 (commented):
lgtm

PR #4136, opened 2026-08-29
fix(web-ng): make the TLS policy control reachable for every transport provider

mfreeman451:
Saving a UniFi Protect or Axis credential rule failed **every time** with `Invalid TLS policy`. These rules have never been saveable through the UI.

Refs #4209.

## Cause

Three sites disagreed about one descriptor key:

| Site | Behaviour |
|---|---|
| render (`:768`) | `descriptor_values(auth_descriptor, "tls_policies")` — defaults to `[]` |
| gate (`:803`) | `show_tls_policy? = rule_controls["transport"] and tls_policies != []` |
| validate (`:1352`) | `enum_param(params, "tls_policy", @tls_policies, ...)` — **unconditional** |

`unifi-protect` sets `transport: true` but declares no `tls_policies`, so the select never rendered, the browser submitted nothing, and the save failed.

Raising the reader's default would have changed nothing: `IntegrationDescriptor` **always materializes** `tls_policies`, normalizing an absent manifest key to `[]`, so `Map.get(descriptor, key, default)` never reaches the default. Core already reads `[]` as "any policy permitted" (`CredentialIntegration.allowed?/2`) — the form was the odd one out.

## Changes

- `effective_tls_policies/1` gives the render and validate sites **one shared** notion of "not narrowed", so they cannot drift apart again.
- `tls_policy_param/1` mirrors the existing `ssh_host_key_policy_param/2` and falls back to the resource default, so an unrendered control can never make a form unsaveable again.
- `show_tls_policy?` now tests `ssh_host_key_policies == []`. An empty `tls_policies` no longer distinguishes an SSH transport, and without this an SSH key-pair credential started offering a TLS policy — a regression the first commit introduced and the third fixes.
- The generic `controller_host` control no longer hardcodes UniFi copy, now that `netbox` also enables it.

## Verification

Run against a migrated `srql-fixtures` scratch database — **22 tests, 0 failures**.

Both fixes are load-bearing under mutation:

| Mutation | Result |
|---|---|
| Restore the original `tls_policies != []` gate | 2 failures |
| Drop the `ssh_host_key_policies` guard | 1 failure (pre-existing SSH test) |
| Neither | 22/22 pass |

Every existing fixture declared `tls_policies`, which is exactly why this shipped unnoticed; the new `api_key` method reproduces the real manifest shape.

> Note: the first run exited **0 having executed zero tests** — web-ng skips DB tests unless `SERVICERADAR_REQUIRE_DB_TESTS=1`. Worth knowing before trusting a green local run.

Not run: `make test` (full Bazel sweep).

## Merge order

Merge this **before** #4134 — that branch touches the same LiveView in different regions and should rebase onto this.


mfreeman451 (commented):
lgtm

PR #4138, opened 2026-08-29
WIP: preserve chart first-drag range selection

mfreeman451:
Preserves the two-file chart first-drag selection patch recovered from a temporary worktree.\n\nVerification: the focused Vitest file passes 17 tests twice, and the worktree contains no generated files.\n\nThis is deliberately draft and may be superseded by the newer shared chart-range lifecycle already on staging; the PR exists so that comparison can happen without losing the original patch.


mfreeman451 (commented):
lgtm

mfreeman451:
Closing as superseded: staging already has ChartRangeSelectionController covering the binding guard + drag threshold (firstmate sr-4138). Leaving the old patch PR closed so it does not stay as open WIP.

PR #4139, opened 2026-08-29
preserve divergent local USP-02 wire-contract slice

mfreeman451:
Preservation-only draft for a divergent local `usp-02-wire-contract` tip discovered during WIP cleanup.

GitHub already had a different historical branch under the original name, so this exact local tip was pushed separately rather than force-updating either history.

This branch is 1,515 commits behind the current `usp-01-proposal` integration branch and represents an early observation-wire-contract slice: the initial cross-language schema/golden fixtures plus an older OpenSpec state. Current USP work likely supersedes substantial portions.

Do not merge as-is. Use this PR to compare or recover any missing intent, then close it once reconciliation is confirmed.

Preservation SHA: `8d0be018ec77c00c56693cc58d7366be7933b1ff`.

Verification performed during preservation: pre-push gitleaks scanned both commits and found no leaks. The historical tests were not rerun against the current integration branch.


mfreeman451:
Reconciliation is complete. This PR was created only to preserve the divergent local USP-02 wire-contract tip and must not be merged.

The current `usp-01-proposal` contract supersedes this slice through the producer-neutral `record.proto` and expanded `sweep.proto` schemas, their registered cross-runtime corpora, and the current OpenSpec ownership. Merging this historical branch would duplicate protobuf/generated symbols, restore an incompatible `MtrTraceEventV1` field assignment, and lose the current retryable-disposition semantics.

No remaining invariant needs to be recovered from this draft. Closing as superseded while retaining `archive/usp-02-wire-contract-local-20260829` at preservation SHA `8d0be018ec77c00c56693cc58d7366be7933b1ff`.

Issue #4209, opened 2026-08-31
Integration plugin credential delivery: finish and verify the demo fixes


mfreeman451:
## Summary

Every credential-backed integration plugin in `demo` is failing, and **none of the failures is a missing credential**. All 13 rows in `platform.network_credential_secrets` are already `source_type = internal_encrypted`, the Helm chart carries zero integration credentials, and the AWX controller reports `AWX 24.6.1 reachable`. The credentials are in the database exactly as intended; the delivery path around them is broken in five independent places, and **each failure reports as something other than what it is**.

Diagnosis is complete and four branches exist with 12 commits. This issue tracks what remains to finish and verify.

## Root causes (all verified against the live demo DB and the code)

| # | Symptom in the UI | Actual cause |
|---|---|---|
| 1 | `UniFi Protect: 0 cameras, 0 streams` | Wasm sandbox **denied the request**. Stored detail says `collection_error: "host error -2 (http_request)"`; `-2` is `pluginErrDenied`. `allowsHTTPHost` deliberately refuses to expand an `allowed_domains: ["*"]` wildcard to an IP literal and falls to `allowed_networks`, which only the NetBox manifest declared. Controller is `192.168.1.1`. A TLS error would have been `-5`/`-6`, never `-2`. |
| 2 | `Invalid TLS policy` when saving a rule | The TLS control renders only when the auth method declares a non-empty `tls_policies`, but `normalize_rule_params/2` requires the value **unconditionally**. Neither the unifi-protect nor axis manifest declares the key, so the input never rendered and the browser submitted nothing. These rules have **never** been saveable through the UI. |
| 3 | `AWX configuration invalid: api_token is required` | The AWX bridge is a command bridge whose token arrives per dispatch, but any assignment without `action-only:v1` also gets a 60s periodic runner invoking `run_check` with params that hold no token by design. |
| 4 | `NetBox inventory_sync has no sources configured` | Assignment params are literally `{}`, and the manifest declared no `integrations.credential_profiles` at all, so NetBox could not appear in the credential UI. Its API token was designed to sit in plaintext in `plugin_assignments.params` — a direct policy violation. |
[...]

PR #4135, opened 2026-08-29
WIP: document credential management and plugin delivery

mfreeman451:
Preserves the four-commit credential-management documentation and OpenSpec change found only in a local worktree.\n\nThe docs distinguish released and unreleased mechanisms, correct the NetBox credential-rule flow, and record the integration-plugin delivery proposal.\n\nVerification: strict OpenSpec validation and whitespace checks pass. Draft pending review alongside the related implementation branches.

PR #4211, opened 2026-08-31
docs: drop NCO-only framing from validation-runs

mfreeman451:
## Describe your changes

The public Validation Runs page read as an OpenText Network Automation
runbook for a single customer (farm01 / UAL). The API is a general
composite-check re-probe contract: any HTTP client can POST IP + partition
and poll a named check's verdict.

- Rewrite `docs/docs/validation-runs.md` (was `nco-validation-runs.md`) so
  callers, examples, and prerequisites are vendor-neutral.
- Same treatment for Device Facts (`device-facts.md`).
- Isolation verdicts stay as a labeled example of one check's decision
  table, not the API's own contract.
- Redirect `/docs/nco-validation-runs` and `/docs/nco-device-facts`.

## Issue ticket number and link

n/a

## Code checklist before requesting a review

- [x] I have signed the DCO?
- [ ] The build completes without errors?
- [ ] All tests are passing when running make test?

Docs-only change. Did not run `make test`.

PR #4213, opened 2026-08-31
docs(credentials): correct the guide against what actually ships

mfreeman451:
`docs/docs/credentials.md` landed in #4135 and is good, but it was written against a moving target: sibling branches shipped underneath it and falsified several statements. This is a **correction pass**, not a restructure. Refs #4209.

## Corrections

**Version markers were stale in the reader's favour.** Boxes said "Not in 1.4.46" while **1.4.49** is current — and `allowed_networks` / `action-only:v1` actually **shipped in 1.4.48**. Verified by walking the tags:

```
v1.4.46, v1.4.47  -> absent
v1.4.48, v1.4.49  -> present on unifi-protect, axis, opentext-nom, awx
```

A reader on 1.4.49 seeing "Not in 1.4.46" concludes it shipped and goes looking. Corrected, and the shipped mechanisms are removed from the unreleased inventory.

**NetBox scope types** said `agent, gateway, partition`. #4134 narrows the profile to `agent` only and adds `IntegrationDescriptor.validate_single_cardinality_scope_types/4`, which rejects any profile pairing a `target_cardinality: single` consumer with a wider scope. The page now documents the constraint and *why*: a NetBox instance is one API endpoint, and a wider scope makes every agent under it walk the same listing and emit a duplicate `snapshot_complete` under the same `source_instance`.

**"There is no environment-variable fallback"** was false for every released version — `fallback_vulncheck_token/0` reads `VULNCHECK_API_TOKEN`, then `SERVICERADAR_VULNCHECK_TOKEN`, then app config. True only with #4134. Now states the 1.4.49 behaviour plainly and marks the removal.

## Two claims caught by review and fixed

- *"One assignment per rule per agent is exactly the multiplication this profile has to avoid"* — **inverted**. One assignment per `(rule, agent)` is the intended emission; what the agent-only scope prevents is a rule naming *many* agents.
- *"recorded verbatim in `last_error`"* — `fail_feed` writes `inspect(reason)`, so the field holds the inspected tuple with the message inside it.

## Merge safety

The earlier pass rewrote the CA trust-material box, which **collided** with #4134's rewrite of the same box (`git merge-tree` reported `CONFLICT`). That edit is reverted here so #4134's version wins; the header renumber for that one box is done on #4134 instead.

Verified against the pushed branches:

```
fix/netbox-proxmox-credential-delivery        clean
fix/credential-rule-tls-policy-control        clean
```

## Verification

Every retained assertion re-checked against the real source on the named branch, not inferred. All files ASCII-only.

> The docs build sets `onBrokenLinks: 'throw'` but **not** `onBrokenAnchors`, so a broken in-page anchor warns and still exits 0. Anchors were checked by hand rather than trusted to the build.

## Merge order

Merge after #4136 and #4134 — it describes behaviour those PRs introduce.


mfreeman451:
Rebased onto current `staging` so BazelCI picks up the SweepGroup test-agent helper. The previous run failed `integration_tests_serial_2` / `serial_5` (`agent '<id>' not found`) — same staging gap as #4211/#4212. Docs commits are unchanged.

PR #4223, opened 2026-08-31
fix(core): OpenText credential rules stay Pending; worker calls missing :by_policy

mfreeman451:
## Summary

UAL `ual-nom` stays **Pending** / **Awaiting provisioning** with Run Now disabled. `PluginIntegrationReconcileWorker` discards every pass:

```
ArgumentError: No such read action on resource PluginAssignment: :by_policy
Available: :read, :by_package, :by_agent, :by_edge_principal,
           :all_partitions_for_policy, :by_partition_source_key
```

The provisioner listed assignments with `:by_policy`. That action does not exist. Other policy reconcilers already use `:all_partitions_for_policy`.

## Test plan

- [x] Confirmed on Example CNPG: approved `opentext-nom-inventory` 0.1.4 and 0.1.6, rule `ual-nom` enabled, zero plugin assignments, producer schedules unbound, worker discarded/retryable with the `:by_policy` error
- [ ] After merge: worker logs `Reconciled plugin credential integrations`; rule Runtime becomes On demand/Scheduled; Run Now is clickable


mfreeman451 (commented):
lgtm

PR #4250, opened 2026-09-02
fix(ci): fetch fixture CA bundles over LAN HTTPS

mfreeman451:
## Why

Security audits keep reducing other source hardening to one hole: CI fetches the private fixture CAs over plaintext HTTP. That is TOFU. Scratch images already carry distro public roots, so the fetch can be real TLS if we wrap it in a publicly trusted cert.

The chicken-and-egg only exists if the publisher is terminated by a cert issued by the *same* custom CA. Serving `dgraph-ci-ca` / `srql-fixture-server-ca` over Let's Encrypt is not circular.

## What

- Clients now dial:
  - `https://srql-fixture-ca.carverauto.dev/ca.crt`
  - `https://dgraph-ci-ca.carverauto.dev/ca.crt`
- Caddy stays HTTP as the Envoy backend. Envoy on `lan-shared-gateway` terminates TLS with the existing `*.carverauto.dev` Let's Encrypt wildcard.
- Config rules reject `http://` CA bundle URLs.
- BuildBuddy `task_allowed_private_ips` gains `192.168.6.87/32` so workflow actions can reach the LAN VIP (Calico BGP does not DNAT it to 10.42).

Depends on gitops PR that attaches those hostnames to `lan-shared-gateway` and labels the namespaces. Merge that first so DNS and the HTTPRoutes exist before CI starts using the HTTPS URLs.

## After merge

1. Merge the gitops PR and wait for Argo to program the HTTPRoutes / ExternalDNS.
2. Apply namespace labels if they are not already live (`k8s/srql-fixtures/namespace.yaml`, `k8s/dgraph/ci/namespace.yaml`).
3. Apply the new `k8s/dgraph-ci` Argo Application once.
4. Roll the BuildBuddy executor/workflow values so the LAN VIP allowlist is live before the next HTTPS fetch.

## Tests

- `bazel test //k8s/dgraph/tests:dgraph_endpoint_contract_test`
- `bazel test //k8s/srql-fixtures/tests:ca_bundle_https_contract_test`
- `bazel test //:ci_heavy_gate_contract_test`
- `bazel test //config/manager_validator/rust/tests:validator_integration_test`
- `bazel test //config/manager_validator/go:validator_test`
- `bazel test //config/proto:binding_drift_go_config_test`


mfreeman451:
Gitops companion (merge first): https://github.com/carverauto/gitops/pull/63

PR #4265, opened 2026-09-03
fix(inventory): recover AWX IPs and stop addressless census creates

### LINKED ISSUE #4050: AWX imports IP-less devices: no guard, and the recovery that would fix it has been dead code since 2026-07-12
Found while auditing 156 live IP-less devices on the carverauto/demo cluster. 35 of them are AWX.

## Correction to the obvious reading

The AWX IP-less devices have `first_seen_time` between 2026-07-06 and 2026-07-18, which looks like "it happened once and was fixed." **It wasn't.** `first_seen_time` is frozen at first insert — `device_records.ex:135` does `first_seen_time: prefer_non_nil(existing.first_seen_time, incoming.first_seen_time)`, and the on-conflict update set at `device_writes.ex:756-853` sets `last_seen_time` and `modified_time` but deliberately never `first_seen_time`.

So that date range marks when AWX's **host set stopped growing**, not when the import stopped. Every later sync upserts the same uids and leaves `first_seen` untouched. The path is live.

## The recovery exists and is dead code

Two commits, five days apart:

**`eb8005da22` (2026-07-07)** — "fix(ansible): populate device IP from AWX host ansible_host" added `awx_ansible_host_ip/1` to re-derive the IP when the plugin shipped none. It reads:

```elixir
# device_discovery_ingestor.ex:250-259
with awx when is_map(awx) <- awx_metadata(device),
     raw when is_binary(raw) <- string_value(awx, ["variables"]),
     host when is_binary(host) <- ansible_host_from_variables(raw),
```

**`1856332974` (2026-07-12)** — "fix(awx): minimize inventory discovery metadata" changed what the plugin emits: `-"variables": host.Variables` / `+"ansible_host": ansibleHost`.

The Elixir recovery still looks for `metadata.awx.variables`. The plugin has not emitted that key since. **Nothing failed loudly — the fallback simply never fires.**

Verified in-tree: `"variables"` survives in `go/cmd/wasm-plugins/awx/main.go` only at `:3499`, as the JSON struct tag for *decoding the AWX API response*. The emitted metadata carries `"ansible_host": ansibleHost` at `:3726`.

**The value the recovery wants is present under a different key.** This looks like a one-line fix, but see the caveat below.

## No guard has ever existed

`strong_enough?/1` (`device_discovery_ingestor.ex:229-232`) is the only admission gate on this path. It admits a device carrying **device_id OR ip OR mac OR integration_id**. The AWX plugin always sets `DeviceID` (`main.go:3685`), so the gate can never reject an AWX host regardless of IP. Unchanged since 2026-05-01.

The origin is `main.go:3668-3687`: `ip := ""` and it is only filled when `extractAnsibleHostFromVariables` returns non-empty **and** `isProbablyIP` passes. A host with no `ansible_host`, or whose `ansible_host` is a DNS name, ships `IP: ""`.

Nothing downstream requires an IP. `device.ex:518-521` declares `:ip` with no `allow_nil? false` and no constraint, and the actual writer is the raw-Ecto path AGENTS.md warns about — `device_writes.ex:154-162` `Repo.insert_all(..., on_conflict: ...)`, no Ash changeset, so a guard in an Ash change module would be blind to it by construction.

## Three other IP-less producers, listed to close the audit

These are separate from AWX and mostly intentional, but they write NULL into the same column:

- `device_writes.ex:793` — `WHEN btrim(EXCLUDED.ip) = '' THEN NULL` clears a stored good IP when an incoming **blank** (not nil) arrives. AWX ships nil, so AWX does not clobber; any source shipping `""` does.
- `device_writes.ex:550, :659` — `Map.put(record, :ip, nil)` when a strong-identity device loses an active-IP conflict.
- `agent_links.ex:728` — raw `UPDATE platform.ocsf_devices SET ip = NULL` in the collector-absorption remediation.

## Suggested work

- [ ] **Point the recovery at the key the plugin actually emits.** Read `ansible_host` (falling back to `variables` for older payloads still in flight). Add a test that fails if the plugin's emitted key and the ingestor's expected key ever diverge again — that divergence is the whole bug, and it was invisible.
- [ ] **Decide the policy question the maintainer raised:** should a device with no IP be importable at all? `strong_enough?/1` treating `ip` as one of four alternatives is what permits it. Note this is a real product decision, not an oversight — an AWX host addressed only by DNS name genuinely has no IP to record, and refusing it means losing the inventory row entirely.
- [ ] Clean up the existing rows only **after** the creating path is closed. Deleting first is how a phantom device came back twice.

## Related

Proxmox legacy identifiers over-merging devices across clusters — filed separately.

mfreeman451:
Fixes #4050 (creating paths for AWX recovery and addressless census). Existing IP-less rows are not deleted in this PR.

## What

AWX inventory sync has been importing devices with `ip = NULL` because the Elixir recovery still read `metadata.awx.variables`. The plugin stopped emitting that blob on 2026-07-12 and stamps `metadata.awx.ansible_host` instead. Recovery never fired; nothing failed loudly.

Addressless RFC 5227 ARP probes are a designed observation (MAC, no address). The decoder correctly keeps them. Persisting them as `platform.ocsf_devices` rows was never decided, and they became four of the 156 IP-less devices on demo.

## Changes

- Recover a blank AWX device IP from `metadata.awx.ansible_host` (and `ansible_ssh_host`), falling back to the old stringified `variables` blob for in-flight payloads. DNS names stay IP-less; we never invent an address.
- Add a cross-language contract test that reads `awxHostMetadata` in the Go plugin so the emitted key and the ingestor lookup cannot drift again.
- Add `SourcePolicy.sufficient_to_create?/1` and consult it in SyncIngestor **before** BatchResolver mints a uid. Addressless census probes cannot create a device. A probe that matches an existing device may still enrich it; blank IP is omitted so DeviceWrites does not clear a stored address.
- AWX hosts addressed only by DNS name may still create. That is real inventory with no IP to record.

## Not in this PR

- Deleting the existing 156 IP-less rows. Close the creating path first; next AWX sync should fill IPs where `ansible_host` is an address.
- Other IP-less writers (blank-IP upsert clear, active-IP conflict `nil`, collector-absorption remediation). Follow-up on this branch.

## Test plan

- [x] `mix test --no-start` on the recovery, contract, and SourcePolicy census files (46 tests, 0 failures)
- [ ] Integration wiring tests in `sync_ingestor_mdns_enrichment_test.exs` (need fixture DB)
- [ ] After deploy: confirm new AWX syncs write `ip` when `ansible_host` is an address; addressless census probes do not mint rows. Do not delete existing rows until that is verified.

- [x] I have signed the DCO?


mfreeman451:
Follow-up on this branch: Proxmox, NetBox, and hypervisor enrichment already skip address-less records at the producer. `SourcePolicy.sufficient_to_create?/1` now copies that rule so a slipped payload cannot mint the row they refused.

Still not in this PR: existing-row cleanup, blank-IP upsert clears (designed handoff), active-IP conflict `nil` (identity preservation), and the collector-absorption literal-`agent` remediation.

PR #4268, opened 2026-09-03
fix(netprobe): restore TCP flow attribution across kernel layouts

### LINKED ISSUE #4029: Flow attribution produces no attributed flows: netprobe sees only agent ICMP, and the netprobe host is absent from sampled NetFlow
Diagnosed on the farm01 cluster, 2026-08-25. **Not a regression** — measured over 14 days, it has essentially never worked.

## Symptom

The correlator enriches `platform.ocsf_network_activity` with `ocsf_payload -> event_type = 'attributed_flow'`. It produces almost nothing:

| window | flows | attributed |
|---|---|---|
| last 6 h | 882 | **0** |
| last 14 d | ~68,000 | **3** |

The agent→core half is **healthy** and should not be suspected: `platform.flow_process_attribution_current` holds 134 rows from `agent-alma-test01`, refreshing continuously (newest `updated_at` advanced during the diagnosis). The ack/quarantine delivery queue is working. The failure is entirely in the 5-tuple join.

### (a) netprobe attributes only ICMP, and only from the agent itself

Every one of the 134 rows looks like this:

| proto | local_ip | remote_port | comm |
|---|---|---|---|
| 1 (ICMP) | `0.0.0.0` | 0 | `serviceradar-ag` |
| 58 (ICMPv6) | `::` | 0 | `serviceradar-ag` |

That is the ServiceRadar agent's own sweep pinging the subnet. There is **not one TCP or UDP row** — even though the agent's own gRPC connection to agent-gateway is TCP over a real socket and should appear.

A 5-tuple join keyed on local/remote IP + port can never match `0.0.0.0:0`, so these rows are structurally unjoinable regardless of what the flow side contains.

### (b) The netprobe host is absent from the sampled NetFlow

One sampler in the window (the site edge router; its public address is redacted here because this repository is public). It exports flows for plenty of `192.168.1.x` hosts — `.224`, `.59`, `.101` and others — and **zero** involving `192.168.1.171`, which is the only host running netprobe.

Worth recording how that was checked, because the first check could have lied: `src_endpoint_ip::text = '192.168.1.171'` would be a false negative if `inet` rendered a `/32` suffix. Re-checked with a prefix match; `inet::text` renders bare here, so the zero is real.

## Net effect

The correlator has ICMP-with-port-0 on one side and router-observed TCP/UDP on the other, with no shared endpoint. There is nothing to join. Both gaps must close before a single flow can be attributed — fixing either alone changes nothing.

## Proven vs inferred

**Proven** (queried directly, read-only):
- 0 attributed flows in 6 h, 3 in 14 days
- all 134 attribution rows are ICMP/ICMPv6 from `serviceradar-ag` with `local_ip 0.0.0.0`/`::` and `remote_port 0`
- `192.168.1.171` appears in zero sampled flows as source or destination
- one sampler in the window
- the agent→core path is live and current

**Inferred, needs host confirmation:** *why* netprobe sees no TCP/UDP. The ICMP-only pattern fits socket-lifecycle eBPF programs failing to attach or producing no events, while raw/ICMP sockets remain visible by a different path. That should be confirmed on the host before any code change — check whether the TCP/UDP socket-lifecycle programs are attached, and whether `flow_table` and the attribution maps have entries.

## Suggested next steps

1. On the netprobe host: confirm which eBPF programs are attached and whether the attribution maps hold non-ICMP entries. Prior art worth re-reading first — a `&key` from the caller's frame passed into an `#[inline(never)]` helper made every map insert fail **silently** while the program verified, loaded and ran. Symptom was zero entries with a healthy-looking probe.
2. Decide whether flow attribution is expected to work at all in this topology. It structurally requires a host that both runs netprobe **and** whose traffic traverses a sampled export path. If no such host exists in a deployment, the feature is inert there by design, and that deserves to be visible rather than silently producing zero.
3. Consider an observability gate: today "the correlator ran and matched nothing" and "the correlator never ran" look identical from the outside. A countable metric for candidate rows considered vs. rows stamped would have surfaced this in minutes rather than by manual query.

## Related

- `openspec/changes/refactor-netprobe-onto-generic-addon-contract` task 5.6 — a separate audit concluded the acked-relay move should **not** proceed; it would not have helped here, since this failure is downstream of delivery.
- Two real defects in the same path, being fixed separately: a false-ack into a non-durable 100-entry gateway buffer, and head-of-line blocking on core's singleton `StatusHandler`.

mfreeman451:
Closes #4029

## Summary

- select the supported 8-byte or 16-byte `inet_sock_set_state` tracepoint layout from live tracefs metadata
- validate every consumed `sock_common` field against kernel BTF and report TCP attribution unavailable for unsupported layouts
- harden owner cleanup, bounded producer telemetry, exact-first correlation ranking, and the operator diagnostic path
- add a production-object loopback lifecycle test and bump the netprobe add-on to `0.2.53`

The observed enterprise-Linux failure was an `inet_sock_set_state` common-header mismatch: the host exposed a 16-byte header while the object decoded the original 8-byte layout, shifting every semantic field and exiting before tuple/ring submission.

## Verification

Passed:

- repository-wide `make test`: **209/209 test targets** ([BuildBuddy invocation](https://carverauto.buildbuddy.io/invocation/60f5165e-d86b-43b9-9d71-a9e412f37f17))
- exact `//rust/netprobe:loaded_tcp_attribution_test` build after declaring its test-only dependency (3,066 actions)
- `cargo test -p serviceradar-netprobe --lib kernel_layout::tests` (7 tests)
- `cargo test -p serviceradar-netprobe --lib attribution::tests` (25 tests)
- Linux-target `cargo check` and `cargo clippy -D warnings` for netprobe
- `cargo fmt --package serviceradar-netprobe -- --check`
- `bazel --batch build -c opt --config=remote //rust/netprobe/ebpf:netprobe_ebpf_object`
- `bash scripts/check-native-addon-version-bumps.sh github/staging HEAD`
- `openspec validate harden-flow-attribution-pipeline --strict`
- two implementation review rounds resolved; final scoped review clean

The broad local run used immutable Ubuntu snapshot packages matching the existing `rules_rs` SHA pins because the fork still references a rolling-archive URL that now returns 404. No build-infrastructure workaround is included in this product branch.

Still environment-gated:

- executing (rather than building) the privileged production-object test requires a Linux runner with `CAP_BPF`, `CAP_PERFMON`, and `CAP_NET_ADMIN`
- scratch-CNPG and live rollout/canary tasks remain explicitly unchecked in the OpenSpec ledger.


mfreeman451:
BazelCI after the dependency fix successfully built the workspace, passed all 200 selected unit targets, and passed the `god_view_elk_scene_acceptance` target. The job then failed in the integration preflight because the shared `sr_core_template` contains migration `20260901140000`, which is not present on current `staging` (`e94f09c4a8`). The preflight explicitly classifies this as an ahead-of-checkout shared-template condition; it is not a failure in this branch.

CI invocation: https://carverauto.buildbuddy.io/invocation/e68f3f10-6cbe-4331-9cd2-c5acfc30eeef
Passing local full unit invocation (209/209): https://carverauto.buildbuddy.io/invocation/60f5165e-d86b-43b9-9d71-a9e412f37f17

mfreeman451:
Latest BazelCI failure isolated and fixed in 4f4308c424.

Root cause: //rust/netprobe:loaded_tcp_attribution_test requires CAP_BPF, CAP_PERFMON, and CAP_NET_ADMIN, but the generic database integration wildcard forces tests onto the capability-free workflow runner. The test now remains an explicit privileged target and is marked manual so ordinary wildcard lanes do not misclassify a missing worker prerequisite as a product failure. BazelCI still names the target explicitly in the build step, preserving compile coverage.

Local verification:
- full make test: 209/209 passed — https://carverauto.buildbuddy.io/invocation/03107ec8-f81a-4951-b3cd-39c6e10733f2
- workflow contract: passed — https://carverauto.buildbuddy.io/invocation/46a2d91a-494f-4121-be68-5c0e3b12fd43
- native add-on build gates: 9/9 passed — https://carverauto.buildbuddy.io/invocation/ab209834-87db-4f6f-a350-3be886a2d519
- native add-on version-bump guard: passed against 4f4308c424fc6fabbcbed36146a28dc20637f6d0

The broad local lint contract also completed Go, Swift, Rust, compile/format, and Credo stages; it stopped only when the freshly updated Hex advisory database reported new repository-wide Ash-family advisories outside this PR diff.

PR #4301, opened 2026-09-04
test(core): release advisory locks between ingestion chunks

mfreeman451:
## Summary

- run the large ingestion release gate outside the test-wide SQL sandbox transaction
- assert that transaction-scoped advisory locks are released after ingestion
- preserve the serial lane and disposable database lifecycle

## Root cause

The Armis identifier ownership trigger takes one `pg_advisory_xact_lock` per device. The release gate wrapped all 50 ingestion chunks in one rollback-only sandbox transaction, retaining tens of thousands of locks until PostgreSQL exhausted shared lock memory. Production commits each chunk independently; `sandbox: :unboxed` makes the release gate exercise that transaction lifetime.

Observed failure: [BuildBuddy invocation](https://carverauto.buildbuddy.io/invocation/042b7f7c-a82c-4080-b129-a6ab31ef4700).

## Verification

- regression RED: 10-device focused scratch-database run observed retained advisory locks and failed the new assertion without `sandbox: :unboxed`
- GREEN: same fresh scratch database lifecycle, 442 migrations, and 10-device workload passed with 1 test / 0 failures after the tag
- `mix format --check-formatted test/release_gates/large_ingestion/results_router_release_gate_test.exs`
- `git diff --check`

The full 50,000-device `//elixir/serviceradar_core:large_ingestion_release_gate` is intentionally verified by the in-cluster LargeIngestionGate workflow.


mfreeman451 (commented):
lgtm

PR #4302, opened 2026-09-04
netprobe: a filter string can no longer kill the process, or mean the wrong port

mfreeman451:
Five findings against the capture filter compiler, all of which matter more now that [#4298](https://github.com/carverauto/serviceradar/pull/4298) made it reachable from an operator-supplied string over IPC — and eventually from an RPCAP client's own expression.

Four of the five came from the original S1 review. **The fifth came from adversarially reviewing my own first attempt at this fix, which was wrong** — see below.

## 1. Deep nesting aborted the process, and bounding *nesting* was not enough

The obvious shape:

```
"(".repeat(10_000) + "tcp" + &")".repeat(10_000)
  -> fatal runtime error: stack overflow, aborting
```

is fixed by a depth guard in `parse_not`, which every recursive **parser** path passes through.

**That left the crash fully reachable by a shape the guard cannot see.** `parse_expr` consumes an `and`/`or` chain in a *loop*, so `tcp or tcp or ...` never drives the depth counter above 1 — while the tree it builds leans left one level per operator. Two unguarded recursions then walk that spine: `Codegen::expr`, which recurses into its left operand for every `And`/`Or`, and the derived `Drop` for `Box<Expr>`.

Measured through the real entry point (`request::validate` → `filter::compile`) on a tokio worker's 2 MiB stack:

```
n=  1000 bytes=   7003 -> capture filter is too complex: jump offset exceeds 255
n=  5000 bytes=  35003 -> ABORT (before this change)
n= 19000 bytes= 133003 -> ABORT (before this change)
```

133 KB, against a 4 MiB frame cap. `and`, `||` and chains of distinct primitives all abort at the same order of magnitude. A stack overflow is not a catchable error, so none of this was "a bad filter is rejected" — it was netprobe dying and taking eBPF flow attribution and the device census with it, from one string typed into a capture form.

So **terms** are charged as well as depth, bounding the tree by construction whatever its shape. Neither existing ceiling helps, because both are raised *after* codegen has already recursed: `resolve()`'s "jump offset exceeds 255" and `Program::new`'s instruction cap. Same input, after:

```
n= 19000 bytes= 133003 -> capture filter is too complex: filter has too many terms (limit: 1024)
n= 50000 bytes= 350003 -> capture filter is too complex: filter has too many terms (limit: 1024)
```

The limit costs nothing real. Classic BPF jump offsets are 8-bit, so this compiler already refuses an `or` chain at about **29 terms** — everything the term limit rejects was unrepresentable anyway; it just rejects it before the recursion instead of after. A test *derives* the longest compilable chain and asserts the limit sits above it, so the guard cannot start rejecting filters that would have worked.

## 2. Port literals used Rust's radix rules instead of libpcap's — and the first fix had a hole

libpcap runs the literal through `strtol(..., 0)`, so C's base prefixes apply. Confirmed by md5 of `tcpdump -dd` on 4.99.4 / Linux 6.8:

```
port 010   == port 8    (710dc899d6d5)
port 0x50  == port 80   (c6c5ba8da5de)
```

Reading `010` as decimal was the dangerous half: **both** compilers accept it and they mean different ports, so an operator copying a working tcpdump filter got a capture of the wrong traffic with nothing to indicate it.

The hole: the radix is sniffed off the **raw** string while the digits are parsed afterwards, and `from_str_radix` accepts its own leading sign — so the two halves disagreed with each other. `+010` does not start with `0`, took the decimal arm, and meant port 10 while `010` meant port 8: two spellings an operator reads as one literal, resolving to different ports. `0+70` was worse, taking the octal arm with digits `+70` and compiling to **port 56** — which my own commit message had claimed could not happen.

Every sign is now refused, in every radix. That is a *narrowing* — libpcap's `strtol` would take some of these — and the direction is deliberate: an operator who gets an error retypes the filter; one who silently captures a port they did not write does not know to.

## 3. The CIDR prefix length had the identical divergence, one function away

`net 10.0.0.0/010` is `/8` to tcpdump and was `/10` here, which silently **narrows** the filter — a capture that quietly misses traffic the operator asked for. Both now share one `parse_c_integer`.

## 4. `src`/`dst` before a primitive with no direction was accepted, then silently dropped

libpcap answers `syntax error` to `src ip` and `dst tcp`, verified against 4.99.4. Accepting them and discarding the qualifier compiled a strictly **wider** filter than the operator wrote — `src ip` captured both directions — and it looks like it worked.

The regression test now names every primitive in the guard's match arm. An earlier version omitted `icmp6`, which meant that one could have gone back to widening with the whole suite still green.

## 5. The differential test's floor was replaced badly the first time

The exact product of the two input lengths catches a `continue` slipping into the loop, but it is *derived from those lengths* and so cannot catch either input shrinking — which the old `compared >= 250` genuinely did catch. 250's real weakness was conflating the two inputs: half the corpus with twice the expressions passed it, and a failure named neither side. Both are now asserted separately, alongside the product.

## How these were found

Findings 1 and 5 exist because I put my own first version of this change through an adversarial review before opening it — five reviewers on separate failure dimensions, each finding verified by three independent agents trying to *refute* it. The stack-overflow gap was reported by four of the five independently and survived 0/3 refutation attempts, with a reproduction through the production entry point on the right stack size. My commit message's headline claim was false, and I would have shipped it.

17 findings survived verification, 11 were refuted. Everything confirmed is fixed here.

## Verification

- 395 netprobe unit tests pass; `cargo clippy --all-targets` clean
- `bazel test //rust/netprobe:netprobe_test //rust/afpacket/... //rust/netprobe:live_capture_ipc_build_test` — all pass
- add-on version gate passes at 0.2.60
- The flat-chain abort reproduced and then re-run against the fix, through `request::validate` on a 2 MiB stack — the same path an IPC capture request takes


mfreeman451 (commented):
lgtm

Issue #4167, opened 2026-08-30
Expose sweep diagnostics through SRQL and ServiceRadar MCP


mfreeman451:
## Problem
When a device is targeted by multiple sweep groups and/or agents, it is difficult to determine which scanner profile, sweep group, vantage point, and requested protocols produced each result. In particular, devices imported from CSV and tagged `rids` are configured for TCP ports `3001,443,4502`, but live results currently show ICMP-only availability with no way to inspect the compiled sweep configuration or per-execution protocol outcome through ServiceRadar MCP.

## Requested enhancement
Expose read-only sweep diagnostics through both SRQL and the ServiceRadar MCP surface, including:

- sweep groups: target query, partition, assigned agent/vantage point, scanner profile, interval, enabled state, and effective ports/modes;
- scanner profiles and the effective compiled configuration delivered to each agent;
- sweep group executions and their timestamps/status;
- per-device, per-agent latest availability, retaining sweep-group/execution context;
- per-device/per-execution host results with requested modes, observed modes, response, scanned/open ports, timestamps, agent/vantage point, and sweep-group identity;
- overlap/last-writer diagnostics showing when multiple groups or agents target the same device.

## Acceptance criteria
1. An operator can identify every sweep group and scanner profile that targets a device.
2. Results can be queried historically and grouped by agent/vantage point and sweep group without one result overwriting another.
3. Requested ICMP/TCP modes can be distinguished from observed protocol outcomes, including closed/no-response TCP ports.
4. The same information is available through documented SRQL entities/fields and ServiceRadar MCP tools, with normal partition/authorization boundaries.

This is a read-only observability/debugging enhancement; it must not expose credentials.

mfreeman451:
Already implemented on staging. Closing as done.

Landed:
- https://github.com/carverauto/serviceradar/pull/4276 — persist attempted port coverage and durable sweep attribution
- https://github.com/carverauto/serviceradar/pull/4303 — SRQL sweep diagnostics entities
- https://github.com/carverauto/serviceradar/pull/4314 — device_sweep_overlap declared-vs-observed diagnostics

Write path, rollup (`platform.sweep_coverage_daily`), and SRQL/MCP entities are on staging. Reopen only if a specific query still fails against live data.

PR #4303, opened 2026-09-04
feat(srql): sweep diagnostics entities and a gated entity-permission fix

mfreeman451:
Read-only SRQL surface for sweep diagnostics — the query half of #4167. The collection half
(scanned-port coverage, the daily rollup, the retention guard) merged earlier in #4276.

## What this adds

Five SRQL entities, all gated:

| entity | reads | answers |
|---|---|---|
| `sweep_groups` | `platform.sweep_groups` | which groups exist, their target query, partition, agents, profile, interval |
| `sweep_profiles` | `platform.sweep_profiles` | scanner profiles: ports, modes, timeouts |
| `sweep_executions` | `platform.sweep_group_executions` | run history with status, timings, agent, config version |
| `sweep_results` | `platform.sweep_host_results` | per-host results incl. `scanned_ports` vs `open_ports` |
| `sweep_coverage` | `platform.sweep_coverage_daily` | per-day coverage that outlives the 7-day raw retention |

`scanned_ports` minus `open_ports` is what makes a refused TCP port distinguishable from one that was
never attempted — the question #4167 opens with.

## Authorization

This also closes a **live authorization bypass** that predates this work.

`EntityAccess.extract_entity/1` resolved the queried entity with an anchored `~r/^in:(\S+)/`, but the
Rust parser accepts `in:` at any token position and assigns unconditionally on every `in` token — so
the LAST one wins. Two consequences, both verified against the compiled NIF:

- `limit:1 in:devices` extracted `"limit:1"`, matched no permission, and ran **ungated**.
- `in:logs limit:1 in:merge_audit` authorized against `logs` while executing against `merge_audit`.
  A caller holding only `observability.logs.view` could read a `devices.view` entity.
- `IN:devices` (capitalized key) fell through to passthrough entirely.

The gate now resolves the entity exactly as the parser does — last token wins, case-insensitive key.
Verified by running `Native.parse_ast` side by side with `extract_entity/1` across 2- and 3-token
queries in every ordering.

The five new entities are mapped to a new `networks.sweeps.view` permission (`@all_roles`, matching
`devices.view` / `composite_checks.view` — read-only diagnostic data, no credentials).
`settings.networks.manage` was deliberately not reused: it is a manage permission, so binding read to
it would conflate read with write and lock out read-only viewers.

## Not in this PR

`device_sweep_overlap` and `sweep_compiled_config` are the remaining two entities and follow
separately. Their SQL has been validated against a real migrated Postgres but no code is written yet.
Per-entity `stats:` and the 90-day query-window lift for `sweep_coverage` also follow.

## Verification

- `cargo test` 648, clippy and fmt clean.
- web-ng db-free tier 1541 tests, `entity_access_test.exs` 19.
- Every fix in this branch was proven with a red-then-green cycle, and the authorization fixes
  additionally against the real compiled parser rather than by reading it.

Refs #4167


mfreeman451:
Added `aadc3ce07f` after review: **a row-level authorization bypass this PR would otherwise have shipped.**

`sweep_profiles.admin_only` is a genuine read restriction in Ash — `authorize_if expr(admin_only == false); authorize_if is_admin()` (`sweep_profile.ex:101-104`) — respected by the settings page via `Ash.read(SweepProfile, scope: scope)`. The SRQL entity had no such filter, so once it was gated on `networks.sweeps.view` (`@all_roles`), a **viewer could read admin-only scanner profiles** through `in:sweep_profiles` — data denied on every other surface.

SRQL's raw-SQL path has no actor context, so per-caller row filtering isn't possible there; the only precedent for row-level authz is `dashboards`, which is special-cased out of the SQL path into Ash. So the fix is an unconditional `admin_only = false` predicate in `build_query`, making the entity a strict subset of what any caller may already read. `admin_only` is now rejected as a filter token rather than silently ignored, and the catalog description states the exclusion so an empty result reads as "restricted" rather than "no such profile".

Tradeoff worth knowing: admin-only profiles are invisible via SRQL **even to an admin**, who must use the settings UI. Routing the entity through Ash instead would preserve that at the cost of SRQL's filter/stats surface — happy to revisit if you'd rather have it the other way.

Found by review, not by the implementer or by the 28-agent review of the commit that introduced it.

Issue #4277, opened 2026-09-03
Shared sr_core_template can ratchet ahead of staging, blocking LargeIngestionGate until an unrelated branch merges


mfreeman451:
## Summary

An unmerged branch's migration can be applied to the shared `sr_core_template`, and because the template only ratchets forward, **every subsequent push to `staging` fails `LargeIngestionGate` until that branch merges**. On 2026-09-03 this held `staging` red across at least six consecutive merges.

The failure is not a broken migration, a broken test, or a flaky fixture. It is an ordering hazard in the shared-template design: the template's contents can get ahead of `staging`, and nothing prevents that or recovers from it automatically.

## What it looks like

`prepare_template` fails before any test runs:

```
Error: template sr_core_template is AHEAD of this checkout: 1 migration(s) are applied
that this branch does not contain.

Extra applied versions: 20260902100000

The template is shared and only ratchets forward, so a branch behind another clones a
FUTURE schema. `mix ecto.migrate` cannot see this: it reports migrations pending from
this checkout, not migrations already applied to the template that are absent from this
checkout.

Merge or rebase onto the branch that added those migrations. Re-provisioning will NOT
help -- a fresh clone of the same template reproduces it exactly.
```

The diagnostic itself is good — it names the condition, the version, and says re-provisioning will not help. The problem is that a `staging` merge has no action available to it: the fix is on someone else's branch.

## This instance

| | |
|---|---|
| Contaminating migration | `20260902100000_add_identity_reconciliation_runs.exs` |
| Added by | `b76ddb8d7d` on `feat/4229-identity-reconciliation-diagnostics` |
| Which was | **PR #4270 — open and unmerged while `staging` was failing** |
| Window | 2026-09-03 ~05:11Z through ~13:15Z, when #4270 merged |

`staging` merges that failed, all with byte-identical errors:

| staging commit | PR | invocation |
|---|---|---|
| `9d4a5c6edd` | #4264 | [`bd3b860e`](https://carverauto.buildbuddy.io/invocation/bd3b860e-7fed-43d4-95c6-9701e6a321ae) |
| `e94f09c4a8` | #4262 | [`ba6684d4`](https://carverauto.buildbuddy.io/invocation/ba6684d4-7389-4270-8b54-93036e08dc8d) |
| `93ccb42f72` | #4251 | [`e1688ff9`](https://carverauto.buildbuddy.io/invocation/e1688ff9-4c00-49b2-b8df-670c2cc81f7e) |
[...]

PR #4314, opened 2026-09-05
feat(srql): device_sweep_overlap declared-vs-observed diagnostics

mfreeman451:
Task 4 of issue #4167. Adds the sixth sweep diagnostics entity: a view-backed
`device_sweep_overlap` pairing every sweep-group **declaration** against the coverage rows
actually produced, so `relationship = 'declared_not_observed'` names a device a group claims
to sweep but has never reported on.

Stacked on #4303 (merged), which added the first five entities and the `networks.sweeps.view` gate.

## Shape

Follows the `addon_fleet` view-backed pattern rather than the five typed-table entities: the
view is deliberately absent from `schema.rs` and read via `sql_query` with a single `Jsonb`
payload. `execute()` delegates to `to_sql_and_params()` so the SQL that runs is the SQL
`translate` returns — Diesel does not rewrite `?` for Postgres and `?` is a valid Postgres
operator, so a divergence surfaces as a syntax error at the *following* token. That class of
bug shipped in five entities before this one (#4285).

## The SQL was validated before it was written

Running the planned shape against a real migrated Postgres found four defects, three worse
than hypothesized. The worst: a group declaring one scanned target and one silent target
produced **zero rows for the silent one** — `declared_not_observed`, the entity's entire
purpose, could not find the thing it exists to find. Also fixed: a text-vs-CIDR comparison,
NULL join keys yielding ~50% false alerts at 144k scale, and an ownership flag reading TRUE
when no row existed.

The corrected form is ~15x faster: 4.6s vs 70.7s on a full scan, 1.8s vs 13s on the
`declared_not_observed` headline query, both at 144k declared x 142k observed rows.

## Verification

Against a clone of the current CI template (449 migrations) plus a synthetic fixture
(RFC 5737 documentation IPs only):

```
declared_and_observed | dev-fixture-a | 192.0.2.10 | exact
declared_not_observed | dev-fixture-b | 192.0.2.20 |          <- the silent target
observed_not_declared | dev-fixture-z | 192.0.2.99 |          <- negative control
```

Made falsifiable rather than trusted: deleting the declaration in a transaction moved
`declared_not_observed` 1 -> 0, rollback restored it to 1.

- `cargo clippy --all-targets -- -D warnings`: clean
- `cargo test --lib`: 678 passed, 0 failed (12 for this entity, incl. the `?`-placeholder guard)
- `entity_access_test.exs`: **red 3 failures -> green 0**. Without the gate the failure names the
  hole directly: `[{"device_sweep_overlap", :passthrough}, {"sweep_overlap", :passthrough}]` —
  an unmapped entity is `:passthrough`, i.e. readable without `networks.sweeps.view`.
- VizMeta checked column-for-column against the view's actual SELECT: 26/26, same order.
- Scratch database dropped and re-verified absent; shared template untouched at 449.

## Note for the reviewer

This carries a new migration, so its first BazelCI run ratchets the shared `sr_core_template`
forward and will block every open PR that lacks it — see #4277. That is the mechanism that
blocked this stack earlier today, not a defect in this branch.


mfreeman451:
## Review response — three findings fixed, all proven against real Postgres

An adversarial review found one **security** issue of the same class as the one just closed in
#4303, plus two correctness/usability defects. All three are fixed in `a989859581`.

### 1. `admin_only` scanner profiles leaked through the view

The view projected `sp.name AS scanner_profile_name` and `sg.profile_id` through a plain
`LEFT JOIN platform.sweep_profiles` with **no `admin_only` guard**. `admin_only` is a row-level
Ash read restriction (`sweep_profile.ex`: `authorize_if expr(admin_only == false)`), and
`networks.sweeps.view` is `default_roles: @all_roles` — so every authenticated user could read
the identity of a profile `in:sweep_profiles` correctly hides from them.

Closing this bypass on `sweep_profiles` and leaving it open on a view that joins the same table
would have been no fix at all.

Proven red-then-green against a clone of the current CI template, with a synthetic fixture
(RFC 5737 IPs, one `admin_only` profile and one ordinary profile as negative control):

```
BEFORE (unmasked)
 declared_not_observed | dev-admin-side    | Group Admin    | Restricted Deep Scan | profile_id_exposed=t
 declared_not_observed | dev-ordinary-side | Group Ordinary | Ordinary Scan        | profile_id_exposed=t

AFTER (masked)
 declared_not_observed | dev-admin-side    | Group Admin    | <masked>             | profile_id_exposed=f
 declared_not_observed | dev-ordinary-side | Group Ordinary | Ordinary Scan        | profile_id_exposed=t
```

The columns are **masked, not the row dropped**: "group G declared this device and never swept
it" is the alert this view exists to raise, and it is equally the operator's business whether or
not the profile behind it is admin-only. The negative control confirms the mask is not blanket.

### 2. The default sort buried the entity's only alert

`declared_not_observed` rows carry a NULL `last_seen_at` **by construction** — never observed is
what makes them alerts. The default `ORDER BY last_seen_at DESC NULLS LAST` therefore sorted
every alert behind every ordinary row. At the scale this view targets (144k x 142k) the ordinary
prefix exceeds `max_cursor_offset = 100_000`, so the alerts were not merely buried, they were
**unpageable**.

Verified on real data:

```
OLD default: observed_not_declared, declared_not_observed, declared_not_observed
NEW default: declared_not_observed, declared_not_observed, observed_not_declared
```

An explicit `sort:` from the caller still replaces the default entirely (covered by a test).

### 3. `last:24h` was silently ignored

Neither the SQL builder nor the guard read `plan.time_range`, so a time window returned all-time
rows with nothing to say it had been dropped. Now **refused** rather than applied — binding a
window to `last_seen_at` would delete every alert row, so there is no honest meaning to give it.

### Also

Corrected a load-bearing comment describing `ocsf_devices_unique_active_ip_idx` as
`(partition, ip)`. It is unique on `ip` alone, so the partition scope is there for correctness
(an unscoped lookup can resolve into another partition), not to enable the index.

### Verification

- `cargo test --lib`: **680 passed, 0 failed** (was 678; +2 for the new sort and time-range behavior)
- `cargo clippy --all-targets -- -D warnings`: clean
- Scratch database dropped and re-verified absent; shared template untouched at 449 migrations

### Deliberately not fixed here

The review also raised NULL-agent coverage rows suppressing per-agent `declared_not_observed`
(real, but the fix changes the compat predicate's semantics and wants its own change), and the
materialized-CTE fences limiting scaling headroom. Both are follow-ups rather than scope creep on
this PR — happy to file them.

mfreeman451:
## Follow-up `52f1eabf93` — the fix in `a989859581` was reachable around, plus the OpenSpec work

The previous commit fixed the default sort **in the Rust query**. It did not hold, because
nothing in that layer is what the UI sends.

### 1. The visual builder reinstated the buried ordering on every query

`catalog.ex` declared `default_sort_field: "last_seen_at"` for `device_sweep_overlap`, and
`Builder.maybe_add_sort/3` emits a `sort:` token for whatever it finds there. An explicit
`sort:` **replaces** the alert-first default entirely — which is stated in the previous commit's
own review response — so every query the builder produced re-created exactly the condition that
commit set out to remove: `declared_not_observed` rows behind a prefix that at this view's scale
exceeds `max_cursor_offset`, so they cannot even be paged to.

Blanking the field is the precedent already set by `vulnerability_advisories`, whose catalog
entry carries the same rationale in a comment: *"The Rust query has a stable compound default...
Leaving this blank keeps the visual builder from replacing it with a weaker single-column sort."*

### 2. Found by the test written for (1): no sweep alias reached the catalog at all

`@entity_aliases` had no sweep entries. Every alias the SRQL parser accepts and `EntityAccess`
gates — `sweeps`, `sweep_group`, `scanner_profiles`, `scanner_profile`, `sweep_profile`,
`sweep_execution`, `sweep_group_executions`, `sweep_result`, `sweep_host_results`,
`sweep_coverage_daily`, `sweep_overlap` — missed it and fell through to the synthesized fallback
in `entity/1`:

```
sweep_overlap -> %{id: "sweep_overlap", label: "Sweep_overlap",
                   default_sort_field: "timestamp", filter_fields: []}
```

So the builder emitted `sort:timestamp:desc` against entities with no `timestamp` column, and
offered no filter fields. This affects the five entities from #4303 as well as this one. I fixed
the whole map rather than one line of it — the defect is identical per alias and the fallback is
silent, which is why it survived two reviews. Say the word if you'd rather I split the other ten
out.

### 3. OpenSpec

- `specs/srql/spec.md`: the "Sweep overlap and last-writer diagnostics" requirement said the
  scanner profile "SHALL be reported", which the `admin_only` masking in `a989859581`
  deliberately does not do. Added scenarios for the masking (row kept, profile nulled), the
  alert-first ordering, and the time-window refusal.
- `tasks.md`: ticked 3.7 and recorded what `device_sweep_overlap` completed under 5.3/5.4. The
  shared 3.x bullets stay unticked because `sweep_compiled_config` still owes them — added a note
  saying so, since six of seven entities have landed and an unticked shared bullet was reading as
  "nothing done".
- `openspec validate add-sweep-diagnostics-srql-entities` — valid.

### 4. Cookbook

Added the "where do groups overlap" recipe (task 5.3), covering the three `relationship` values
and the two behaviors an operator would otherwise misread: a NULL `scanner_profile_name` means
**restricted**, not absent; a time window is **refused**, not ignored.

### Verification

- `//elixir/web-ng:unit_tests` — **8/8 shards pass** on RBE
  ([invocation](https://carverauto.buildbuddy.io/invocation/209c9c29-e692-42db-b20c-9a0815161a54))
- `mix format --check-formatted` and `mix credo --strict` — clean (1295 files, no issues)
- SRQL builder tests 37/37; the new alias test fails red on `sweep_overlap` before the fix
- Three `NotificationsLive.ProviderUploadTest` failures appear under the **local** full async
  suite only. Reverting this commit's code and re-running the same suite still fails there, the
  file passes 23/23 in isolation, and the Bazel shard that owns it (`unit_tests_phoenix_other`)
  passes — pre-existing local flake, not this change.

mfreeman451:
## BazelCI red for a shared-fixture reason, not a reason in this diff

`//elixir/serviceradar_core:migrate_template` fails before any migration on this branch runs
([invocation](https://carverauto.buildbuddy.io/invocation/73be89f3-9666-462f-b4e8-ca2abebec5c1)):

```
empty database; applying schema baseline
[SchemaBootstrap] Applying schema baseline from .../priv/repo/baseline/platform_schema.sql (1771 statements)
** (Postgrex.Error) ERROR 42501 (insufficient_privilege) must be owner of extension timescaledb
```

The template database is empty and the role applying the committed baseline does not own the
`timescaledb` extension. That is the fixture rebuild in flight (#4277, #4313), not this change:

- The failure is in the **baseline**, so this branch's `add_device_sweep_overlap_view` migration
  is never reached.
- #4315, whose diff shares nothing with this one, fails with the **identical** signature on
  `migrate_run` — and its actual suite passes 207/207 before the provisioning step dies.
- This PR's last two commits touch only `catalog.ex`, the MCP cookbook, a test file, OpenSpec
  markdown and `AGENTS.md`. No migration, no schema, no Rust.

Every other check passes (15/15). Locally, the command CI runs for the non-fixture graph is green
on RBE from this exact tree:

```
bazel test -c opt --config=remote //... --test_tag_filters=-integration_test,-acceptance_test
Executed 30 out of 216 tests: 216 tests pass.
```
([invocation](https://carverauto.buildbuddy.io/invocation/508f48f5-aaa0-41c3-9c48-94349b86ce82))

Not retrying BazelCI here — the template rebuild is being handled in #4313, and retrying against
a half-rebuilt fixture is what makes that worse. Needs a re-run once #4313 lands.

PR #3597, opened 2026-08-17
feat: durable execution pipeline (usp-01)

mfreeman451:
Migrated from Forgejo.

Original Forgejo pull request: #4685
Original author: @mfreeman451
Original URL: https://code.carverauto.dev/carverauto/serviceradar/pulls/4685
Original created: 2026-07-21T15:11:43Z

---

## Slice 1 — proposal: edge sweep/MTR result data plane

Base of the `unify-sweep-results-proto` stack. OpenSpec proposal that replaces
the JSON-in-gRPC sweep/MTR result path with a layered, durable, native-protobuf
data plane:

scanner -> crash-safe segmented agent spool -> bounded protobuf frames over
independent mTLS gRPC lanes -> stateless gateway -> JetStream PubAck ->
EventWriter -> CNPG.

Key decisions: a sweep is a logical stream of independently-useful microbatches
(transport EOF != completion); durable acceptance only on JetStream PubAck;
bulk/interactive/recovery traffic separated; MTR summaries stay small sweep
observations while full traces use their own bounded protobuf events; idempotent
partitioned CNPG projection with AckExplicit; `network_scope_id` is a site
identity + fairness key, not tenancy. Reconciles edge-architecture,
observability-signals, mtr-diagnostics, cnpg, nats-*, age-graph, sweeper,
sweep-jobs, agent-config/connectivity; archives the superseded
add-bulk-payload-pipeline; revises dependent add-adhoc-network-scan to canonical
events.

Files: proposal.md, design.md, tasks.md, and the affected-capability spec deltas
under specs/. Addresses review #4674.

PR #4316, opened 2026-09-05
fix(edge): make the lane ledger outlive its transport, and fail closed without one

mfreeman451:
Stacked on #3597 (`usp-01-proposal`). Implements criterion **(i) RESTART OVERLAP** of `unify-sweep-results-proto` task 3.3. **3.3 stays unchecked** — see "What this does not close".

## The defect

A lane was one `:one_for_all` restart unit over `{NATS connection, PublisherPool}`. `PublisherPool.init/1` builds an *empty* window, so any restart handed the lane its **full grant back** while requests admitted under the previous accounting were still in flight on the previous connection. Old and replacement requests together could exceed the grant. `:one_for_all` narrowed that interval; it did not close it, because supervisor restarts are ordered but not instantaneous.

## The shape of the fix

The lane splits into a **stable accountant** and a **replaceable transport generation** under `LaneSupervisor`'s `:rest_for_one`, **accountant first**. The order is the invariant, and the two directions are deliberately asymmetric:

| | what happens | why |
|---|---|---|
| **transport dies** | accountant is *before* it, so it is untouched | every charge survives; the replacement inherits the *remaining* capacity, never a fresh grant |
| **accountant dies** | transport is *after* it, so it is terminated first | a fresh accountant has an empty ledger, and an empty ledger beside live send capability is the same defect wearing a different hat |

The fresh accountant additionally starts **closed**: `admit/4` returns `{:error, :no_transport}` until a new generation registers, which under `:rest_for_one` cannot happen until the previous send capability is gone.

## Fencing ends attempts, never reservations

Each attempt records the transport `generation` it was issued on (a `reference()`, not a counter — a counter restarts with its process, so a stale generation would compare equal to a fresh one and fencing the old would fence the new). When a generation dies, `PublishWindow.fence_generation/2` ends its attempts while **keeping their reservations charged**.

That asymmetry is the whole point: generation death proves the request cannot complete *on that transport*, and proves **nothing** about whether the bytes reached the broker. Affected publications become idle-but-charged and may retry on the charge they already hold. Only a validated resolving PubAck releases credits.

## Tests

Mutation testing found a real gap: the fail-closed guard could be **deleted** and the accepting-generation reset removed, and every test still passed — because every pool helper registers a transport in setup, so the closed state was the one state no test ever saw. Four tests, deliberately not using that helper, now kill both mutants:

- a fresh accountant refuses until a transport registers
- a refusal for want of transport charges nothing
- a dead transport closes it again while the charge from before the death survives
- a replacement re-opens it on the **remaining** capacity rather than a fresh grant

## What this does not close

- **3.3 stays unchecked.** This invariant is proven against the **serial** publisher that exists today. An invariant exercised only serially is not an invariant under concurrency, and the asynchronous pipeline it must also hold under is not built. That is the next PR (async worker pool + out-of-order PubAck accounting).
- **A dead owner's reservation stays charged.** `tasks.md` previously claimed (i) would free it. It does not: fencing fires on the death of a *transport generation*, not of an owner, so an owner that dies while its transport stays healthy leaves a charged reservation with no attempt against it. That is deliberate — owner death is not evidence the record was not published — but it is a real retention gap, and bounding it needs evidence that the specific request terminated. That is 3.5's correlation work, not a supervision change here. The doc correction is in this PR.


mfreeman451 (commented):
Requesting changes for one state-machine gap and the new compiler warnings.

1. **[P1] `register_transport/2` does not enforce the new generation bound or registration order.** `PublisherPool.handle_call({:register_transport, pid}, ...)` (lines 264-273) unconditionally mints a generation, makes it accepting, and `Map.put`s it into `transports`. It neither rejects a third known generation nor proves the registering pid is the current live registrar. That contradicts the new requirement at `openspec/changes/unify-sweep-results-proto/specs/ingestion-routing/spec.md:405-406` (at most one accepting and one draining generation) and the state comment at `publisher_pool.ex:241-243`. Under rapid replacement, registrations and monitor `:DOWN`s come from different senders, so multiple replacements can be accepted before older downs are handled. A late registration from an already-dead registrar can also overwrite a newer accepting generation; its immediate `:DOWN` then sets `accepting` to nil while the newer live generation remains only in `transports`, leaving the lane closed with no registrar that will re-register. Please make registration an explicit bounded transition (including stale/dead caller handling) and add tests for a third registration and a late dead registration after a live replacement.

2. **[P2] This branch introduces compiler warnings, including unreachable test clauses.** The focused Bazel run emits:
   - `publisher_pool.ex:404`: the catch-all `handle_info/2` is separated from the earlier clause at line 361 by `caller_down/2`.
   - `publisher_pool_test.exs:361-365`: the newly added nested `eventually/2` duplicates the already-complete helper at lines 326-335. Because `describe` does not create a module namespace, the later clauses are module-level, and the compiler reports both as unable to match. The new tests currently happen to call the earlier helper, so the mutation proof works, but the duplicate helper is dead code.

3. **[P3] Supervision docs still describe the replaced topology as current.** `LaneSupervisor` still has the heading “Why `:one_for_all`” and says “under `:one_for_all`, a crash in either child restarts both” at lines 5-16, although the implementation is now accountant-first `:rest_for_one`. `PublisherSupervisor` likewise says connections are started before pools at line 16, now the opposite of the invariant. These comments are likely to mislead the next supervision change.

Verification performed against commit `9ebb0aadf7354cea1c7d72642c4d95a16ff71f44`:

- `bazel --output_base=/tmp/fm-sr-usp50-review/bazel test -c opt --config=remote //elixir/serviceradar_core:unit_tests_serviceradar_edge //elixir/serviceradar_agent_gateway:unit_tests_serviceradar_agent_gateway` — both targets passed.
- Mutant removing the explicit closed-state guard — edge shard failed three new assertions (`:generation` instead of required retryable `:no_transport`) on both attempts.
- Mutant retaining a dead generation as accepting — edge shard failed the two close-on-death/replacement assertions on both attempts.
- `openspec validate unify-sweep-results-proto --strict` — passed.

The principal architecture is otherwise sound for the serial publisher: accountant-first `:rest_for_one`, generation-specific attempt fencing, and idle-but-charged reservations preserve the frame/byte ledger; task 3.3 correctly remains unchecked pending the async worker/out-of-order PubAck work, and the dead-owner reservation is accurately left open for 3.5 correlation.

mfreeman451:
All three addressed on `2df987077f`. Thank you — P1 was a real reachable state, not a hypothetical one, and finding it turned up a second shape of the same wedge that the review did not name.

**[P1] Registration is now an explicit bounded transition, and `accepting` is derived.**

The root cause was that `accepting` and `transports` were two separately-assigned facts that could disagree. Rather than guard each site, `accepting` is now **derived** from the retained set at every mutation (`recompute_accepting/1`) — so it cannot name a generation that is not retained, and cannot be nil while a live one remains. Both wedge shapes become unrepresentable rather than merely fixed.

The transition itself, in order:

1. **Reap first.** Any generation whose registrar is already gone is fenced exactly as its `:DOWN` would fence it (attempts ended, reservations kept). This is what makes the bound meaningful: it is measured against generations that are genuinely live, so a replacement is never refused merely because the VM has not delivered a notification yet.
2. `{:error, :dead_registrar}` when the registering pid is not alive.
3. Idempotent for a pid already registered — returns its existing generation, since minting a second reference for one lifetime would consume the bound with a generation no death can clear.
4. `{:error, :generation_limit}` at `@max_generations` (2 = one accepting + one draining).

The refusal reaches `LaneTransportRuntime.Registrar`, which now `{:stop, {:transport_registration_refused, reason}}`s instead of `badmatch`ing — so the generation is retried under the supervisor's restart intensity rather than being admitted past the bound.

**The second wedge, same cause, in `fence/2`.** It set `accepting: nil` whenever the accepting generation died. Whenever another generation was still retained and alive, that closed the lane with send capability it refused to use — and since a registrar registers exactly once from `init/1`, nothing would ever re-open it. It now re-derives, falling back to the newest remaining generation and reaching nil only when none remains (which is the fail-closed state a fresh accountant already starts in).

**Tests: five, each proven to fail against the unfixed code.** I mutated the specific guard each one covers and confirmed it dies, rather than asserting they cover it:

| mutant | test that died |
|---|---|
| drop the `Process.alive?` guard | `an ALREADY-DEAD registrar is refused, and does not displace a live generation` — `left: {:error, :dead_registrar}` / `right: {:ok, #Reference<...>}` |
| raise `@max_generations` to 999 | `a THIRD live generation is refused, not absorbed` |
| `fence/2` closes instead of re-deriving | `when the ACCEPTING generation dies the lane falls back to one that is still live` |

Plus `registering the SAME process twice returns one generation, not two`, and `a DEAD generation is reaped to make room, and its charge survives the reaping` — the latter kills both generations and registers a third *without waiting for either `:DOWN`*, so it only passes if registration reaps them itself, and it asserts the first generation's charge is still held afterwards.

The spec is updated to state the enforcement rather than leave the bound as an assumption (`ingestion-routing`, "A publisher lane loses its accounting"): registration refusal, dead-registrar refusal, derived-accepting, and the fallback. `openspec validate unify-sweep-results-proto --strict` passes.

**[P2] Both warnings gone.** `handle_info/2`'s catch-all and the `handle_call/3` clauses were separated by private helpers; helpers moved below. The duplicate `eventually/2` at `publisher_pool_test.exs:361-365` is deleted — you were right that `describe` creates no namespace, so those clauses were module-level dead code. The focused Bazel run is now warning-free.

**[P3] Four moduledocs corrected, one more than reported.** `LaneSupervisor` and `PublisherSupervisor` as you found them, plus two you did not name that made the same claim:

- `PublishWindow`'s "what this still does not bound" described a replacement pool starting with a full grant — and named the same reservation-freeing behaviour `tasks.md` was already corrected for claiming.
- `JetStreamPublisher`'s settle path said the same. What can now reach that branch is a superseded reservation, or an accountant that died — which under `:rest_for_one` took its transport with it.

Each says what stayed open, too: an owner that dies mid-request still leaves a charged reservation (3.5's correlation work), and the restart invariant is proven only against the serial publisher.

Verification on `2df987077f`: `//elixir/serviceradar_core:unit_tests_serviceradar_edge` and `//elixir/serviceradar_agent_gateway:unit_tests` pass with no warnings. Task 3.3 stays unchecked; the async worker pool and out-of-order PubAck accounting are the next PR, stacked on this one.

mfreeman451:
Housekeeping note, so the force-push here is not a surprise: `5a5431d3b2` accidentally carried two files from the next slice (`publish_pipeline.ex` and its test) into this PR — I ran `git add -A` on this branch while those files were untracked in the worktree. They are removed and the branch is force-pushed to `975e6dc941`, which is the same three commits plus the flake fix, with the flake fix now containing only the one test file it describes.

Nothing already reviewed changed. The diff you reviewed on `2df987077f` is untouched; `975e6dc941` adds only `publish_window_test.exs`.

The async worker pool and out-of-order PubAck accounting are now open as #4317, stacked on this branch. Task 3.3 stays unchecked there, with the reason recorded in `tasks.md`: the implementation is complete but nothing offers to a pipeline yet, since the gateway's per-lane session is 3.1's and `publish_record/2` has no production caller either.

mfreeman451:
**BazelCI is red here, and it is not this diff. Diagnosis from the invocation, not from a nearby guess.**

[Invocation `5a6f0ce9`](https://carverauto.buildbuddy.io/invocation/5a6f0ce9-3334-4b66-a6db-e04a8512e763) fails in the **template preflight**, before any of this branch's targets run:

```
//elixir/serviceradar_core:migrate_template   FAILED
  empty template; applying schema baseline
  ** (Postgrex.Error) ERROR 42501 (insufficient_privilege)
     must be owner of extension timescaledb
     ... SchemaBootstrap.apply_baseline!/2
```

That is BazelCI's preflight running `//elixir/serviceradar_core:migrate_template` **against the shared `sr_core_template` from a branch** — the thing `AGENTS.md` forbids and that #4310 (`fix/ci-shared-template-branch-writes`) fixed on `staging` at 2026-09-05T01:50Z.

**`usp-01-proposal` does not have that fix.** It is 49 commits behind `staging`, and `git merge-base --is-ancestor` confirms the merge commit is absent. The two preflights differ exactly where it matters:

```diff
-  preflight="$(bazel run ... --//build:run_id=$ID //rust/integration-db:prepare_template)"
-  case "$preflight" in
-    *"migration(s) pending"*) bazel test $FLAGS //elixir/serviceradar_core:migrate_template ;;
+  preflight="$(bazel run ... --//build:template_authority=true --//build:run_id=$ID //rust/integration-db:prepare_template)"
+      bazel run ... --//build:template_authority=true ... //rust/integration-db:reset_template
```

So every PR into `usp-01-proposal` currently runs the pre-#4310 preflight, which writes the shared template from a branch. The timeline matches: the last BazelCI success on a `usp-01-proposal` PR was 2026-09-04T18:57Z (#4133), #4310 landed on `staging` at 01:50Z, and `staging`'s own BazelCI was flapping across that window (`b885fa474b` fail 03:45, `6f140861ee` fail 04:18, then `7d4c29b4ff` 04:53 and `25ff48d4e0` 05:07 green once it had the fix). This run was 05:47Z, concurrent with the trunk `LargeIngestionGate` on `fe4ae8cee6`.

**Nothing in this branch touches migrations, the baseline, or database configuration**, and I did not run `migrate_template` myself — BazelCI's own preflight did, per the config on the base branch. The remedy is to merge `staging` into `usp-01-proposal` (the recurring maintenance PR: #4297, #4292, #4246), which is not a change this PR can or should make.

**What was verified locally instead**, on this exact tree:

- `bazel test -c opt --config=remote //... --test_tag_filters=-integration_test,-acceptance_test` — **224/224 pass** ([invocation](https://carverauto.buildbuddy.io/invocation/1dc8100f-b52f-4255-9320-51a88f0d5c37))
- `./scripts/elixir_quality.sh --project elixir/serviceradar_core --lint-only` — format clean, `credo --strict` clean over 2411 files
- `openspec validate unify-sweep-results-proto --strict` — valid
- no compiler warnings on the edge or gateway shards

All 11 GitHub check-runs on this commit are green; BazelCI is the only red, and it is red for the base branch's preflight rather than for anything here.

mfreeman451:
Update on the BazelCI failure above: the fix is #4318, which merges `staging` into `usp-01-proposal`.

`usp-01-proposal` predates the whole shared-template fix chain (#4310, #4313, plus `6f140861ee` and `7d4c29b4ff`), so its BazelCI still runs `migrate_template` against `sr_core_template` from a branch. After that merge, the **BazelCI action contains no template calls at all** — `prepare_template` / `migrate_template` / `reset_template` appear only under `LargeIngestionGate`, behind `--//build:template_authority=true`. The trunk lifecycle becomes the single writer and the PR gate stops touching shared state.

The restack was verified rather than assumed, against the merged base:

- this branch's six commits rebase onto it with **no conflicts**
- #4317 then rebases onto this branch with no conflicts
- `bazel test //... --test_tag_filters=-integration_test,-acceptance_test` on the full restacked stack: **226/226 pass** ([invocation](https://carverauto.buildbuddy.io/invocation/efbf67bb-d562-4d0a-ba59-a817ed6f7239))

Nothing is force-pushed here yet — restacking before #4318 merges would put 49 `staging` commits into this PR's diff. Once it lands, this branch and #4317 rebase onto the updated base and the BazelCI failure goes with it.

mfreeman451:
Restacked onto `usp-01-proposal` now that #4318 has landed (`495974517b`), which brings in the shared-template fix chain. The BazelCI failure reported above should not recur: after that merge, the BazelCI action contains no template calls at all — `prepare_template` / `migrate_template` / `reset_template` run only under `LargeIngestionGate`, behind `--//build:template_authority=true`.

#4318 going green itself settled the open question about which config BuildBuddy uses: it runs the PR's **merge** config, not the base branch's. That is why #4318 could pass a gate its own base was failing, and why this branch picks up the fix immediately rather than needing it on the base first.

Scope is unchanged — 6 commits, 15 files. None of the 49 `staging` commits are in this diff; they arrived via the base.

Verified on `4fc3250b12` before pushing:

- rebase applied all six commits with **no conflicts**
- `bazel test //... --test_tag_filters=-integration_test,-acceptance_test` — **226/226 pass** ([invocation](https://carverauto.buildbuddy.io/invocation/4743ffb0-c344-4f9e-9e09-9a263a9a094f))

PR #4317, opened 2026-09-05
feat(edge): async publish pipeline and out-of-order PubAck accounting

mfreeman451:
Stacked on #4316 (`usp-50-lane-accountant`), which is itself stacked on #3597 (`usp-01-proposal`). **Review #4316 first** — this branch contains it.

Implements the remaining half of `unify-sweep-results-proto` task 3.3(c): asynchronous pipelined publication under the hard window, and out-of-order PubAck accounting. **3.3 stays unchecked** — see "Why 3.3 is still unchecked".

## What was actually missing

Every part existed except the thing that drives them:

| module | does | status before |
|---|---|---|
| `PublishWindow` | bounds admission by frames, bytes, PubAck deadline | landed |
| `PublisherPool` | owns that bound per traffic class | landed |
| `ResolvedPrefix` | advances a watermark across out-of-order outcomes | landed, **unwired** |
| `JetStreamPublisher.publish_record/2` | admit → request → settle, **inline** | landed |

Because `publish_record/2` does all three inline, exactly one frame was ever on the wire per calling process. That is why `PublisherPool` had to say the window *"bounds concurrent admissions but NOT concurrent publications"*. With `PublishPipeline` the two coincide: a worker holds its credits for exactly as long as its request is outstanding, so what the grant bounds is the number of requests on the wire.

## Why each worker admits for itself

This is the load-bearing decision, and the obvious alternative is wrong.

A dispatcher that admitted centrally and handed reservations to workers would break the fencing invariant the pool is built on. The **owner** of an attempt is the process that called `admit/4` — taken from the call's `from` precisely so no caller can name a different process — and only the owner may `settle/3` or `attempt_failed/2`. A handed-off reservation leaves the dispatcher owning a request it did not issue and cannot report on.

So the pipeline hands out **work**, never reservations. Each worker owns its attempt from admission to settlement, which carries task 3.3's post-handoff fencing criterion (ii) into the concurrent case *unchanged* rather than re-arguing it.

## One pipeline per class, never per lane

`nats-tenant-isolation` forbids the installation creating "a durable, connection, **process**, account, or physical stream per network scope, agent, producer assignment, run/execution, output contract, package, or logical partition". A prefix tracker is per **spool lane** — so the tempting shape, a process per lane holding its own tracker, is exactly the prohibited one.

Trackers are therefore **data** in one per-class process, keyed by lane. `open_lane/3` takes a key and starts nothing; a test asserts the pipeline's links are unchanged across opening lanes. Their number is bounded by `:max_lanes` and refused past it, so "data rather than a process" does not quietly become unbounded retention instead.

## What resolves, and what only caps

Conservatively, because the two errors do not cost the same — a sequence wrongly resolved lets the prefix advance past a record that may never have been published, and eventually authorises reclaiming customer data that was never delivered.

```
{:ok, pub_ack}     -> ACCEPTED_AUTHORITATIVE   resolves
{:error, :poison}  -> REJECTED_PERMANENT       resolves (PROVEN poison only)
any other error    -> REJECTED_RETRYABLE       caps the prefix
the worker CRASHED -> REJECTED_RETRYABLE       caps the prefix
```

A worker crash is deliberately in the last row rather than treated as a missing outcome: a process can die *after* its request reached the socket, and its reservation stays charged in the pool because owner death is not termination. That retention is 3.3's known open gap, owed to 3.5's correlation work — this module does not paper over it by pretending the sequence resolved. `{:error, {:derivation, _}}` only caps too: a local bug is not a broker verdict, and nothing about it proves the record can never be accepted.

## Nothing here republishes

Backpressure is a refusal (`:queue_full`) and nothing else. Requeueing internally was the first design and was wrong twice: a capacity refusal redispatched immediately is a hot loop against a saturated window, and every other failure may have reached the socket — so republishing it is a decision about a source sequence this module holds no state to make. Same reason `JetStreamPublisher` returns a retryable error rather than retrying.

## Both closure criteria, now under concurrency

This is the gap #4316 named as its own limit. Serially, "old and replacement requests together cannot exceed the grant" was a claim about a *single* request, because one caller could hold exactly one.

- **(i) restart overlap** — a generation dies with **four** requests on the wire. Every charge survives, and the replacement transport is refused for want of capacity rather than handed a fresh grant. `max_inflight` is set *above* the grant on purpose, so what holds the fifth publication back is the lane's remaining capacity and not the pipeline's own cap — with the two equal the queue would have absorbed it and the assertion would have passed without the grant being consulted.
- **(ii) post-handoff fencing** — a retry offered by a genuinely separate process while an attempt is in flight is refused, on **one** charge. Serially this case could not arise at all.

## The tests can fail

The publisher double goes **through** the window rather than merely blocking — it admits, holds the reservation across a gate the test opens, then settles. A blocking stub would have made every concurrency assertion vacuous: the pipeline would look bounded while the pool bounding it was never consulted. Concurrency is observed from outside, from messages the double sends the test process; nothing asks the pipeline to report its own concurrency, since that number would be wrong in the same way the code would be.

Three mutants, each killed by its own test:

| mutant | test that died |
|---|---|
| `dispatch` ignores `max_inflight` | `max_inflight caps how many publishes are outstanding AT ONCE` |
| `disposition_of/1` resolves every outcome | `a RETRYABLE outcome caps the prefix exactly like a missing one` |
| a dead worker's sequence recorded as accepted | `a worker that DIES leaves its sequence unresolved` |

## Why 3.3 is still unchecked

The implementation is complete, but **nothing offers to a pipeline yet**. The gateway's per-lane session is task 3.1's mTLS bidirectional record RPC, and `JetStreamPublisher.publish_record/2` still has no production caller either — so the whole chain is exercised by tests rather than running. Whether a pre-production implementation discharges (c) is the change owner's judgement, not something to settle here by ticking the box. `tasks.md` records exactly that.

For the same reason the pipeline is **not** in the supervision tree: a process with no runtime reason would violate the repo's own Iron Law. Where it belongs when 3.1 lands — last under `LaneSupervisor`'s `:rest_for_one` — is argued in the moduledoc.


mfreeman451:
**CI coverage note for this PR, since it is not what it looks like.**

This PR targets `usp-50-lane-accountant`, and both real gates are branch-filtered to `staging` and `usp-01-proposal`:

- `buildbuddy.yaml` → `BazelCI` — `triggers.pull_request.branches: [staging, usp-01-proposal]`
- `.github/workflows/elixir-quality.yml` → `mix format --check-formatted` + `mix credo --strict` — same two branches

So what actually runs here is `proto-abi` (path-filtered; this change touches no proto path), `Native add-on version bumps`, and `GitGuardian`. **No tests and no lint.** That is the same gap `elixir-quality.yml`'s own comment records — it "briefly ran by accident while those PRs were grouped in a GitHub stack ... and caught eight credo findings the moment it did — then stopped again when the PRs were unstacked."

Rather than leave that unverified, the full pre-PR suite was run locally against this exact tree (`0908de5e52`):

- `bazel test -c opt --config=remote //elixir/serviceradar_core:unit_tests //elixir/serviceradar_agent_gateway:unit_tests` — **17/17 pass, no compiler warnings** ([invocation](https://carverauto.buildbuddy.io/invocation/525048c7-1e52-4ced-a648-44ac03bf5b3d))
- `./scripts/elixir_quality.sh --project elixir/serviceradar_core --lint-only` — format clean, `credo --strict` clean over 2411 files, no issues
- `openspec validate unify-sweep-results-proto --strict` — valid
- the repo-wide `make test` equivalent is running against this tree as well

Choosing between retargeting this PR at `usp-01-proposal` (full gates, but the diff then shows #4316's commits too) and adding this base branch to the two trigger lists is a call for the change owner, not something to settle inside the PR it affects. Flagging it rather than picking.

mfreeman451:
Restacked on the rebased #4316, which now sits on `usp-01-proposal` post-#4318. Scope unchanged — 5 commits, 8 files.

**The CI caveat from earlier still applies to this PR and is worth restating**, because the check list will look green either way: this PR's base is `usp-50-lane-accountant`, and both real gates are branch-filtered to `staging` and `usp-01-proposal`. So `BazelCI` and `Elixir Quality` do not run here at all. That is not a new consequence of the restack — it is the same gap recorded in the `AGENTS.md` entry this branch adds.

It is also not theoretical. The repo-wide local run on this branch is what caught `//:ci_heavy_gate_contract_test` failing because the new test source was missing from `INTEGRATION_SOURCE_DISPOSITIONS.tsv` — a Bazel target, so no gate on this PR would have run it.

Verified locally on `c7deca3566`, the pushed head:

- rebase applied all five commits with **no conflicts**
- `bazel test //... --test_tag_filters=-integration_test,-acceptance_test` — **226/226 pass** ([invocation](https://carverauto.buildbuddy.io/invocation/4743ffb0-c344-4f9e-9e09-9a263a9a094f))

One follow-up now unblocked by the merge: `ResolvedPrefix` gained `reported_through/2` on `staging` (#4133), a gateway-side retention bound. `PublishPipeline` currently bounds lane *count* via `:max_lanes` but not per-lane retained dispositions, so it should expose that. It could not be written before this restack — the function did not exist on the previous base.

mfreeman451:
#4316 has landed (`eb17818093`), and GitHub retargeted this PR from `usp-50-lane-accountant` to `usp-01-proposal` automatically.

**No restack was needed, and I checked rather than assumed:** the merge-base of this head and the new base is exactly #4316's tip (`4fc3250b12`), so this PR still adds precisely its own 5 commits / 8 files. The new base is also a *pure* merge commit — `git diff 4fc3250b12 usp-01-proposal` is empty — so the tree CI now builds is byte-identical to the one verified at **226/226** locally. Force-pushing a rebase would have churned CI for no change.

**The CI caveat on this PR is now obsolete.** The base is `usp-01-proposal`, which *is* in the trigger lists, so `BazelCI` and `Elixir Quality` run here for the first time — this branch finally gets compiled, tested and linted by CI rather than only locally. The `AGENTS.md` entry this PR adds still stands as guidance for the next stacked PR, which is exactly the situation it describes.

Worth noting what that gap actually cost while it was open: the repo-wide local run is what caught `//:ci_heavy_gate_contract_test` failing on a missing `INTEGRATION_SOURCE_DISPOSITIONS.tsv` row. That is a Bazel target, so no check running on the old base would have caught it — it would have surfaced only now, on retarget.

PR #4318, opened 2026-09-05
Merge staging into usp-01-proposal (unblocks the shared template preflight)

mfreeman451:
Merges `staging` (49 commits) into `usp-01-proposal`, which is what unblocks BazelCI for every PR in the usp-01 stack.

## Why now

BazelCI is failing on PRs into `usp-01-proposal` — see #4316, [invocation `5a6f0ce9`](https://carverauto.buildbuddy.io/invocation/5a6f0ce9-3334-4b66-a6db-e04a8512e763):

```
//elixir/serviceradar_core:migrate_template   FAILED
  empty template; applying schema baseline
  ** (Postgrex.Error) ERROR 42501 (insufficient_privilege)
     must be owner of extension timescaledb
```

That is the pre-#4310 preflight running `migrate_template` **against the shared `sr_core_template` from a branch** — the thing `AGENTS.md` forbids. `usp-01-proposal` predates the whole fix chain, so every PR into it still does this.

The fix chain, all on `staging` and all included here:

| | |
|---|---|
| #4310 `e122cc2a97` | a branch can no longer write its migrations into the shared template |
| #4313 `b885fa474b` | give the shared template database a single writer |
| `6f140861ee` | apply the schema baseline with the privilege the migrations assume |
| `7d4c29b4ff` | replay migrations on an empty fixture database, never the baseline |

After this merge the **BazelCI action contains no template calls at all** — `prepare_template`, `migrate_template` and `reset_template` appear only under `LargeIngestionGate`, gated on `--//build:template_authority=true`. The trunk lifecycle becomes the single writer, which is exactly what #4313 set out to do, and the PR gate stops touching shared state.

## Conflicts

One, in `go.mod`: `usp-01-proposal` carried `github.com/klauspost/compress v1.19.2` in the indirect block while `staging` had promoted it to the direct block. Resolved by taking `staging`'s placement and dropping the duplicate. The resolved file is **byte-identical to `staging`'s `go.mod`** — `usp-01-proposal` had no dependency of its own — and `go mod tidy` is a no-op on it.

## Verification

`bazel test -c opt --config=remote //... --test_tag_filters=-integration_test,-acceptance_test` on the merge result: **226/226 pass** ([invocation](https://carverauto.buildbuddy.io/invocation/9ba0821b-35fa-4e04-9f6e-34dfdfa52f16)).

Once this lands, #4316 and #4317 restack on the updated base and their BazelCI runs stop trying to write the shared template.


mfreeman451:
`proto-abi` was red. Fixed in `d9358c3953`, and the cause is worth stating because it was a merge artifact rather than anything either branch contained.

**The failure:** `protoc-gen-go: program not found or is not executable`, at `Makefile:603` in the Go binding drift guard.

**The cause:** one un-prefixed protoc invocation out of twenty-three. `usp-01-proposal` runs every invocation as

```make
@PATH="$(PROTO_TOOLS_BIN):$$PATH" protoc ...
```

so codegen uses the pinned toolchain `make proto-tools` builds into `bin/prototools`. `staging` has no `PROTO_TOOLS_BIN` at all and invokes a bare `@protoc`, relying on ambient PATH — which is fine there, because `.github/workflows/proto-abi.yml` **exists only on `usp-01-proposal`** and nothing on staging runs this guard.

Both sides touched the same region — staging *added* a `remote_capture` invocation, usp-01 had `rperf` there — and the auto-merge interleaved them:

```make
  @PATH="$(PROTO_TOOLS_BIN):$$PATH" protoc -I=proto -I=. \
      ...
      proto/remote_capture.proto     # staging's new target, took usp-01's PREFIXED header
  @protoc -I=proto -I=. \            # staging's BARE header
      ...
      proto/rperf/rperf.proto        # usp-01's target, now under the bare header
```

Neither side had that. `git merge` reported it as a clean auto-merge.

**Two symptoms, one cause.** On CI, where nothing puts `protoc-gen-go` on PATH, it failed outright — that is the error above. Locally it silently fell back to an ambient `protoc-gen-go`, so `rperf.pb.go` regenerated at `v1.36.11` while everything else regenerated at `v1.36.12`; the drift guard then failed on a *mixed-version tree*, which is the more dangerous half because it points somewhere unrelated to the actual cause.

**Also regenerated `proto/remote_capture.pb.go`.** staging committed it stamped `v1.36.11`. That is not staging being wrong — nothing there regenerates against the `google.golang.org/protobuf v1.36.12` pin in `go.mod`, because the guard does not exist on that branch. Bringing the file onto a branch that *does* enforce the pin means regenerating it. The diff is the version stamp and nothing else:

```diff
-// 	protoc-gen-go v1.36.11
+// 	protoc-gen-go v1.36.12
```

**Verification on `d9358c3953`:**

- `make verify-proto-go` — exit 0, tree clean
- `make verify-proto-elixir` — exit 0, tree clean (CI never reached this step)
- `make proto-lint` — exit 0
- all 23 protoc invocations in `generate-proto` now carry the prefix; zero bare
- `bazel test //... --test_tag_filters=-integration_test,-acceptance_test` — **226/226 pass** ([invocation](https://carverauto.buildbuddy.io/invocation/633171f6-bab5-4ed3-a7f4-9fe568c227ea))

Worth noting for whoever lands the reverse direction: once `usp-01-proposal` reaches `staging`, `staging` inherits both the guard and the hermetic toolchain, and this class of drift stops being invisible there.

PR #4389, opened 2026-09-07
fix(core-elx): drift_mode enum/type mismatch follow-up

mfreeman451:
Follow-up commits landed after PR #4326 merged: pass unknown drift_mode strings through loudly, coerce legacy boolean drift_mode in the anomaly projector.


mfreeman451:
Rebased onto current staging and resolved the conflicts.

Note: both original commits are already on staging in evolved form, so the rebase dropped them as redundant:
- 5848ddcfec (boolean drift_mode coercion + quoted chart defaults) landed as 0bd5544878
- 34026526f0 (pass unknown drift_mode strings through loudly) landed as 33ab3356f8, covered by test "drift_mode rejects boolean true and passes non-boolean values through"

The one surviving delta is the projector comment above normalize_metric_class_value: the inherited wording still claimed a stale row can never fail profile validation, which contradicts the loud-passthrough contract. Restated it (no behavior change).

Issue #4477, opened 2026-09-14
StatefulAlertEngine calls a shard by name after ensure_started, so a just-started shard can return :engine_not_running


mfreeman451:
## Problem

`StatefulAlertEngine` can return `{:error, :engine_not_running}` for a shard it has just confirmed is running.

`fan_out` starts or finds each shard, then throws the pid away and calls the shard by registered name:

- `elixir/serviceradar_core/lib/serviceradar/observability/stateful_alert_engine.ex:171-172`: `with {:ok, _pid} <- ensure_started(shard) do call(shard, {message_tag, records})`; the same shape again at `:202-203`.
- `call/2` (`:325-332`) does `GenServer.call(via_tuple(shard), ...)` and maps an `:exit, {:noproc, _}` to `{:error, :engine_not_running}`.
- `ensure_started/1` (`:338-356`) looks the shard up with `ProcessRegistry.lookup/1` and, when it is missing, starts it with `ProcessRegistry.start_child/1`.

So the call does not go to the process `ensure_started` returned; it goes through a second name lookup. Shards are placed cluster-wide by Horde (per the module's own comments), and when that name resolution lags the start, the call exits `:noproc` even though the shard exists.

## Where it showed up

BazelCI on a PR that touched only Helm files failed `//elixir/serviceradar_core:integration_tests_serial_4` once, and passed on an identical-tree rerun:

```
1) test recovery on an already-terminal alert is an idempotent no-op (no KeyError, no re-resolve) (ServiceRadar.Observability.StatefulAlertEngineTest)
   test/serviceradar/observability/stateful_alert_engine_test.exs:589
   code:  assert :ok = StatefulAlertEngine.evaluate_events([event.("anomaly_open", 0)])
   left:  :ok
   right: {:error, :engine_not_running}
```

ExUnit seed 685945. The test module makes this likely: `setup` calls `TestSupport.drain_stateful_alert_engines()` before and after every test (terminating every registered engine and waiting for the registry to empty), and some tests change `:stateful_alert_engine_shards`, so a test's first `evaluate_events` very often lands on a freshly started shard.

## Why it matters outside tests

The same window exists in production whenever a shard is (re)started: after a crash and Horde restart, or while a rolling deploy moves shards. `evaluate_logs/1`, `evaluate_events/1`, `evaluate_metrics/1` and `resolve_stale_anomalies/…` return `{:error, :engine_not_running}` for that batch even though the shard is up.

## Proposed fix

[...]

PR #4510, opened 2026-09-21
fix(serviceradar_core): stop dropping stateful alert batches on a freshly started shard

mfreeman451:
## Intent

The user asked me to review the 25 newest open GitHub issues on carverauto/serviceradar and pick one to work on. I picked issue #4477: StatefulAlertEngine can return {:error, :engine_not_running} for a shard it has just confirmed is running, which drops the batch. The goal is to stop dropping batches when a shard is freshly started or restarted.

Root cause, confirmed by reading the vendored Horde 0.10 source rather than only the ticket: dispatch_shard/3 and dispatch_resolve_shard/2 threw away the pid returned by ensure_started/1 and then called the shard by its registered name. Horde.RegistryImpl.handle_call({:register, ...}) replies {:ok, self()} right after DeltaCrdt.put, but it writes the name-lookup ETS table only later, from process_diff/2 when it handles the asynchronous CRDT diff. So Horde.Registry.whereis_name/1 can resolve to :undefined for a live, correctly registered, just-started shard; GenServer.call then exits :noproc and the caller got {:error, :engine_not_running}. It was observed as a BazelCI flake on integration_tests_serial_4 on a freshly started shard.

Fix: call the pid that ensure_started/1 resolved. Because that pid can still die between resolution and the call (shard crash, Horde restart, rolling deploy), re-resolve and retry once, then report :engine_not_running. Deliberate decisions: the public contract (:ok | {:error, term()}), the 15s call timeout, and the exit-reason mapping are unchanged; the only behavioural change is that a batch which used to be dropped now gets evaluated, which is intended and is what the issue's proposed fix asks for. No other call sites of the engine exist in the codebase. ensure_started/1 is unchanged, and because Horde's lookup already filters dead pids I did not add a liveness check to it.

Test: added "a shard terminated out-of-band is restarted and takes the next batch" to the existing stateful alert engine test file. Verification: the full test file passed 15 tests / 0 failures against a scratch database on the srql-fixtures CNPG cluster (fresh migrate, pool_size 12; at pool_size 1 the pre-existing shard fan-out test times out on DB checkout, which is environmental, not this change), the new test was stable 3/3 across seeds, and mix format --check-formatted is clean on both files.

Important limitation, included so it is not mistaken for an oversight: I verified that the new test also passes on the unfixed code, so the test does not reproduce the timing race itself. The lag only appears when the registry process is busy enough that the CRDT diff queues behind other messages; it does not reproduce on an idle workstation even with 8 concurrent shard starts. I deliberately rejected two ways to force it: churning the registry to create contention (this permanently bloats Horde's CRDT causal context, which ProcessRegistry's own moduledoc warns about) and poking Horde's private ETS tables (upgrade-fragile). The test therefore pins the deterministic half of the contract, that a terminated shard is restarted and the next batch is evaluated rather than dropped, and the comment in the test states this explicitly. The race itself remains covered by CI as before.

## What Changed

- `StatefulAlertEngine.dispatch_shard/3` and `dispatch_resolve_shard/2` now call the pid that `ensure_started/1` resolved instead of re-resolving the shard's registered name. Horde replies to a registration before it writes the name-lookup ETS row, so the second lookup could exit `:noproc` and return `{:error, :engine_not_running}` for a running shard, dropping the batch.
- `call/2` became `call/3` plus `retry_call/2` and `call_pid/2`: if the resolved pid is gone by the time of the call, the engine re-resolves through `ensure_started/1` and retries once before reporting `:engine_not_running`. The `:ok | {:error, term()}` contract, the 15s call timeout, and the exit-reason mapping are unchanged.
- Added three tests to `stateful_alert_engine_test.exs`: a shard terminated out-of-band is restarted and evaluates the next batch, and `evaluate_events/1` and `resolve_stale_anomalies/3` both succeed against a just-started shard while read-only `:members` calls keep the registry mailbox busy so the lookup lags.

## Risk Assessment

✅ Low: The change is a small, bounded fix that calls the pid ensure_started/1 resolved and retries once only on :noproc (a call that was never delivered, so no double evaluation); the public contract, timeout and exit mapping are unchanged, the fix-round edits are comment rewords plus eventually/2 wrapping of the test's registry reads, and the only residual issue (duplicate shard start in the lookup-lag window) is pre-existing and was already reviewed and ignored by the user.

## Testing

No baseline commands had run for this step. I built a scratch database on the srql-fixtures CNPG cluster and ran the full stateful alert engine test file against the fixed code: all 17 tests passed, and the two new lookup-lag tests took under a second each. I then put only stateful_alert_engine.ex back to the base commit and ran the three new tests against the unfixed dispatch. Both lookup-lag tests failed with {:error, :engine_not_running}. The out-of-band-termination test passed on both versions, which matches the limitation the author declared. For stability, the lag tests passed 5 of 5 seeds on the fixed code, and in a per-test measurement both failed in 8 of 8 runs on the unfixed code. One earlier unfixed run (seed 2) failed overall, but my summary pattern did not capture which test failed, so I did not count it in the per-test figure. Setup note: my first database attempt was wrong. I cloned a leftover scratch database without checking it was migrated (it was empty). I ran `mix ecto.migrate` as the skill file describes, the replay lost its migration lock partway, and a `| tail` pipe hid mix's failing exit status. I dropped that database (it was mine), rebuilt with `mix serviceradar.db.migrate`, and confirmed mix exit 0 and 480 of 480 ledger rows before running any test. No UI surface is involved, so there are no screenshots; the evidence is test transcripts.

- Live validation: ✅ go - 5 of 7 scenarios driven live against the product

| Scenario | Result | Live | Evidence |
| --- | --- | --- | --- |
| An operator&#39;s event batch arrives at a shard that was just started while the registry is busy; the engine returns :ok and evaluates it instead of dropping it | ✅ pass | live | round2-fixed-full-file.txt (test &#39;a batch sent to a just-started shard is evaluated while the registry lookup lags&#39; passes, 12 iterations all :ok); round2-seed-stability.txt FIXED seeds 1-5 |
| The stale-anomaly resolve sweep reaches a just-started shard while the registry is busy; it returns {:ok, count} instead of an error | ✅ pass | live | round2-fixed-full-file.txt (test &#39;a resolve sweep sent to a just-started shard runs while the registry lookup lags&#39; passes, all {:ok, 0}); round2-seed-stability.txt FIXED seeds 1-5 |
| Adversarial: the same lookup-lag scenarios run against the unfixed name-based dispatch must reproduce the dropped batch, proving the regression tests can fail | ✅ pass | live | round2-unfixed-dispatch.txt (left: [:ok, {:error, :engine_not_running}] and [ok: 0, error: :engine_not_running]); round2-unfixed-detection-rate.txt (both tests fail 8/8) |
| A shard terminated out-of-band is restarted by the next batch, the batch produces its alert, and a new live pid is registered under the shard key | ✅ pass | live | round2-fixed-full-file.txt (test &#39;a shard terminated out-of-band is restarted and takes the next batch&#39;); also passes on base code in round2-unfixed-dispatch.txt, consistent with the author&#39;s declared… |
| Existing engine behaviour is unchanged: bucketed alerts, metric baselines, anomaly open/clear, shard fan-out, liveness check and dedup all still work | ✅ pass | live | round2-fixed-full-file.txt: 17 tests, 0 failures, MIX_EXIT=0 |
| The shard pid dies between ensure_started/1 and the GenServer.call; the single retry re-resolves and the batch is still evaluated | ⏸️ untested | no | The window between pid resolution and the call is inside a private function, and nothing available here can force a kill in that window. It would need a test hook in the dispatch path or a tracing-bas… |
| Under BazelCI&#39;s integration_tests_serial_4 lane, where the flake was first observed, the engine no longer drops freshly-started-shard batches | ⏸️ untested | no | The guarded Bazel integration lifecycle may only run in the in-cluster BuildBuddy workflow (typed SERVICERADAR_ENV=ci configuration and private fixture secrets); the project skill forbids running it f… |

<details>
<summary>Evidence: Fixed code: full engine test file, 17 tests / 0 failures</summary>

```text
warning: unique :states [:available, :scheduled, :retryable] is missing incomplete states [:suspended, :executing] which may break uniqueness, use a unique group like :incomplete
  lib/serviceradar/jobs/refresh_trace_summaries_worker.ex:1: ServiceRadar.Jobs.RefreshTraceSummariesWorker (module)

Running ExUnit with seed: 0, max_cases: 1
Excluding tags: [:external, :cluster, :large_ingestion, :benchmark]
Including tags: [:integration]


ServiceRadar.Observability.StatefulAlertEngineTest [test/serviceradar/observability/stateful_alert_engine_test.exs]
  * test fires and resolves alerts based on bucketed counts [L#45]
  * test fires and resolves alerts based on bucketed counts (2310.2ms) [L#45]
  * test fires metric alerts only for sustained baseline violations and recovers [L#133]
  * test fires metric alerts only for sustained baseline violations and recovers (1799.7ms) [L#133]
  * test groups endpoint vulnerability findings by OCSF device uid [L#249]
  * test groups endpoint vulnerability findings by OCSF device uid (993.8ms) [L#249]
  * test causal prediction anomaly routes to ocsf_events and device-grouped alerts [L#325]
  * test causal prediction anomaly routes to ocsf_events and device-grouped alerts (1422.6ms) [L#325]
  * test causal anomaly alerts only on open transitions and resolves on clear [L#465]
  * test causal anomaly alerts only on open transitions and resolves on clear (1696.8ms) [L#465]
  * test anomaly alert liveness check fires, resolves, and discards synthetic artifacts [L#562]
  * test anomaly alert liveness check fires, resolves, and discards synthetic artifacts (2588.9ms) [L#562]
  * test recovery on an already-terminal alert is an idempotent no-op (no KeyError, no re-resolve) [L#594]
  * test recovery on an already-terminal alert is an idempotent no-op (no KeyError, no re-resolve) (1447.1ms) [L#594]
  * test causal anomaly alerts can group by canonical metric tuple without series-key equality [L#715]
  * test causal anomaly alerts can group by canonical metric tuple without series-key equality (947.1ms) [L#715]
  * test central seasonal normal state suppresses matching edge-spike anomaly alert [L#808]
  * test central seasonal normal state suppresses matching edge-spike anomaly alert (1093.2ms) [L#808]
  * test unsupported seasonal metric classes pass through edge-spike anomaly alerts [L#934]
  * test unsupported seasonal metric classes pass through edge-spike anomaly alerts (1223.4ms) [L#934]
  * test resolve_stale_anomalies resolves an open alert whose series went silent [L#1059]
  * test resolve_stale_anomalies resolves an open alert whose series went silent (1601.5ms) [L#1059]
  * test capacity forecast alerts coalesce by resource and resolve on inactive status [L#1155]
  * test capacity forecast alerts coalesce by resource and resolve on inactive status (1626.5ms) [L#1155]
  * test deduplicates repeated event bursts into one active incident and rolls over after cooldown gap [L#1254]
  * test deduplicates repeated event bursts into one active incident and rolls over after cooldown gap (1831.1ms) [L#1254]
  * test fans out across shards so rules in different shards fire concurrently and independently [L#1450]
  * test fans out across shards so rules in different shards fire concurrently and independently (2134.0ms) [L#1450]
  * test a shard terminated out-of-band is restarted and takes the next batch [L#1555]
  * test a shard terminated out-of-band is restarted and takes the next batch (1230.3ms) [L#1555]
  * test a batch sent to a just-started shard is evaluated while the registry lookup lags [L#1650]
  * test a batch sent to a just-started shard is evaluated while the registry lookup lags (766.8ms) [L#1650]
  * test a resolve sweep sent to a just-started shard runs while the registry lookup lags [L#1665]
  * test a resolve sweep sent to a just-started shard runs while the registry lookup lags (855.6ms) [L#1665]

Finished in 26.1 seconds (0.00s async, 26.1s sync)
17 tests, 0 failures
MIX_EXIT=0
```
</details>
<details>
<summary>Evidence: Unfixed dispatch: both lookup-lag tests reproduce the dropped batch</summary>

<code>1) test a batch sent to a just-started shard is evaluated while the registry lookup lags&#10;left: [:ok, {:error, :engine_not_running}]&#10;right: [:ok]&#10;2) test a resolve sweep sent to a just-started shard runs while the registry lookup lags&#10;left: [ok: 0, error: :engine_not_running]&#10;right: [ok: 0]&#10;3 tests, 2 failures (14 excluded)</code>

```text
lib now at base? name-based call present: 1
Compiling 56 files (.ex)
warning: unique :states [:available, :scheduled, :retryable] is missing incomplete states [:suspended, :executing] which may break uniqueness, use a unique group like :incomplete
  lib/serviceradar/jobs/refresh_trace_summaries_worker.ex:1: ServiceRadar.Jobs.RefreshTraceSummariesWorker (module)

Generated serviceradar_core app
Running ExUnit with seed: 0, max_cases: 1
Excluding tags: [:test]
Including tags: [location: {"test/serviceradar/observability/stateful_alert_engine_test.exs", 1555}, location: {"test/serviceradar/observability/stateful_alert_engine_test.exs", 1650}, location: {"test/serviceradar/observability/stateful_alert_engine_test.exs", 1665}]


ServiceRadar.Observability.StatefulAlertEngineTest [test/serviceradar/observability/stateful_alert_engine_test.exs]
  * test fires and resolves alerts based on bucketed counts [L#45]
  * test fires and resolves alerts based on bucketed counts (excluded) [L#45]
  * test fires metric alerts only for sustained baseline violations and recovers [L#133]
  * test fires metric alerts only for sustained baseline violations and recovers (excluded) [L#133]
  * test groups endpoint vulnerability findings by OCSF device uid [L#249]
  * test groups endpoint vulnerability findings by OCSF device uid (excluded) [L#249]
  * test causal prediction anomaly routes to ocsf_events and device-grouped alerts [L#325]
  * test causal prediction anomaly routes to ocsf_events and device-grouped alerts (excluded) [L#325]
  * test causal anomaly alerts only on open transitions and resolves on clear [L#465]
  * test causal anomaly alerts only on open transitions and resolves on clear (excluded) [L#465]
  * test anomaly alert liveness check fires, resolves, and discards synthetic artifacts [L#562]
  * test anomaly alert liveness check fires, resolves, and discards synthetic artifacts (excluded) [L#562]
  * test recovery on an already-terminal alert is an idempotent no-op (no KeyError, no re-resolve) [L#594]
  * test recovery on an already-terminal alert is an idempotent no-op (no KeyError, no re-resolve) (excluded) [L#594]
  * test causal anomaly alerts can group by canonical metric tuple without series-key equality [L#715]
  * test causal anomaly alerts can group by canonical metric tuple without series-key equality (excluded) [L#715]
  * test central seasonal normal state suppresses matching edge-spike anomaly alert [L#808]
  * test central seasonal normal state suppresses matching edge-spike anomaly alert (excluded) [L#808]
  * test unsupported seasonal metric classes pass through edge-spike anomaly alerts [L#934]
  * test unsupported seasonal metric classes pass through edge-spike anomaly alerts (excluded) [L#934]
  * test resolve_stale_anomalies resolves an open alert whose series went silent [L#1059]
  * test resolve_stale_anomalies resolves an open alert whose series went silent (excluded) [L#1059]
  * test capacity forecast alerts coalesce by resource and resolve on inactive status [L#1155]
  * test capacity forecast alerts coalesce by resource and resolve on inactive status (excluded) [L#1155]
  * test deduplicates repeated event bursts into one active incident and rolls over after cooldown gap [L#1254]
  * test deduplicates repeated event bursts into one active incident and rolls over after cooldown gap (excluded) [L#1254]
  * test fans out across shards so rules in different shards fire concurrently and independently [L#1450]
  * test fans out across shards so rules in different shards fire concurrently and independently (excluded) [L#1450]
  * test a shard terminated out-of-band is restarted and takes the next batch [L#1555]
  * test a shard terminated out-of-band is restarted and takes the next batch (1862.1ms) [L#1555]
  * test a batch sent to a just-started shard is evaluated while the registry lookup lags [L#1650]
  * test a batch sent to a just-started shard is evaluated while the registry lookup lags (859.9ms) [L#1650]

  1) test a batch sent to a just-started shard is evaluated while the registry lookup lags (ServiceRadar.Observability.StatefulAlertEngineTest)
     test/serviceradar/observability/stateful_alert_engine_test.exs:1650
     Assertion with == failed
     code:  assert Enum.uniq(results) == [:ok]
     left:  [:ok, {:error, :engine_not_running}]
     right: [:ok]
     stacktrace:
       test/serviceradar/observability/stateful_alert_engine_test.exs:1662: (test)

  * test a resolve sweep sent to a just-started shard runs while the registry lookup lags [L#1665]
  * test a resolve sweep sent to a just-started shard runs while the registry lookup lags (717.7ms) [L#1665]

  2) test a resolve sweep sent to a just-started shard runs while the registry lookup lags (ServiceRadar.Observability.StatefulAlertEngineTest)
     test/serviceradar/observability/stateful_alert_engine_test.exs:1665
     Assertion with == failed
     code:  assert Enum.uniq(results) == [ok: 0]
     left:  [ok: 0, error: :engine_not_running]
     right: [ok: 0]
     stacktrace:
       test/serviceradar/observability/stateful_alert_engine_test.exs:1682: (test)


Finished in 3.8 seconds (0.00s async, 3.8s sync)
3 tests, 2 failures (14 excluded)
MIX_EXIT=2
```
</details>
<details>
<summary>Evidence: Seed stability: fixed 5/5 pass; unfixed runs all fail (seed 2 failed without a captured per-test summary)</summary>

<code>FIXED seed=1..5 mix_exit=0 2 tests, 0 failures&#10;UNFIXED seed=1,3,4,5 mix_exit=2 2 tests, 2 failures&#10;UNFIXED seed=2 mix_exit=2 (summary line not captured by my pattern; not counted per-test)</code>

```text
FIXED seed=1 mix_exit=0 2 tests, 0 failures (15 excluded)
FIXED seed=2 mix_exit=0 2 tests, 0 failures (15 excluded)
FIXED seed=3 mix_exit=0 2 tests, 0 failures (15 excluded)
FIXED seed=4 mix_exit=0 2 tests, 0 failures (15 excluded)
FIXED seed=5 mix_exit=0 2 tests, 0 failures (15 excluded)
UNFIXED seed=1 mix_exit=2 2 tests, 2 failures (15 excluded)
UNFIXED seed=2 mix_exit=2 
UNFIXED seed=3 mix_exit=2 2 tests, 2 failures (15 excluded)
UNFIXED seed=4 mix_exit=2 2 tests, 2 failures (15 excluded)
UNFIXED seed=5 mix_exit=2 2 tests, 2 failures (15 excluded)
```
</details>
<details>
<summary>Evidence: Unfixed detection rate with per-test counting: 8/8 runs, both tests fail</summary>

```text
UNFIXED seed=11 mix_exit=2 | 2 tests, 2 failures (15 excluded) | evaluate_failed=1 resolve_failed=1
UNFIXED seed=12 mix_exit=2 | 2 tests, 2 failures (15 excluded) | evaluate_failed=1 resolve_failed=1
UNFIXED seed=13 mix_exit=2 | 2 tests, 2 failures (15 excluded) | evaluate_failed=1 resolve_failed=1
UNFIXED seed=14 mix_exit=2 | 2 tests, 2 failures (15 excluded) | evaluate_failed=1 resolve_failed=1
UNFIXED seed=15 mix_exit=2 | 2 tests, 2 failures (15 excluded) | evaluate_failed=1 resolve_failed=1
UNFIXED seed=16 mix_exit=2 | 2 tests, 2 failures (15 excluded) | evaluate_failed=1 resolve_failed=1
UNFIXED seed=17 mix_exit=2 | 2 tests, 2 failures (15 excluded) | evaluate_failed=1 resolve_failed=1
UNFIXED seed=18 mix_exit=2 | 2 tests, 2 failures (15 excluded) | evaluate_failed=1 resolve_failed=1
```
</details>
<details>
<summary>Evidence: Helper used to resolve fixture coordinates (contains no credentials)</summary>

```text
# Sourced helper: resolves srql-fixtures scratch-DB coordinates at runtime.
# Holds no credentials; everything secret is read from the cluster into shell
# variables and never printed.
ADMIN_USER=$(kubectl get secret srql-test-admin-credentials -n srql-fixtures -o jsonpath='{.data.username}' | base64 -d)
ADMIN_PASS=$(kubectl get secret srql-test-admin-credentials -n srql-fixtures -o jsonpath='{.data.password}' | base64 -d)
ADMIN_PASS_ENC=$(printf '%s' "$ADMIN_PASS" | jq -sRr @uri)
TLS_SERVER_NAME=srql-fixture-rw.srql-fixtures.svc.cluster.local
CA_FILE="${TMPDIR:-/tmp}/srql-fixture-ca-nm4477.crt"
( umask 077; kubectl get secret srql-fixture-server-ca -n srql-fixtures -o jsonpath='{.data.ca\.crt}' | base64 -d > "$CA_FILE" )
NODEPORT=$(kubectl get svc srql-fixture-rw-ext -n srql-fixtures -o jsonpath='{.spec.ports[0].nodePort}')
DB_HOST=""
for host in 192.168.10.31 192.168.10.96 $(kubectl get nodes -o jsonpath='{range .items[*]}{.status.addresses[?(@.type=="InternalIP")].address}{" "}{end}'); do
  if PGPASSWORD="$ADMIN_PASS" psql \
    "host=$TLS_SERVER_NAME hostaddr=$host port=$NODEPORT dbname=postgres user=$ADMIN_USER sslmode=verify-full sslrootcert=$CA_FILE connect_timeout=4" \
    -v ON_ERROR_STOP=1 -Atc 'select 1' >/dev/null 2>&1; then
    DB_HOST="$host"; DB_PORT="$NODEPORT"; break
  fi
done
admin_psql() {
  PGPASSWORD="$ADMIN_PASS" psql \
    "host=$TLS_SERVER_NAME hostaddr=$DB_HOST port=$DB_PORT dbname=postgres user=$ADMIN_USER sslmode=verify-full sslrootcert=$CA_FILE" \
    -v ON_ERROR_STOP=1 "$@"
}
export_test_env() {
  export SERVICERADAR_TEST_DATABASE_URL="postgres://${ADMIN_USER}:${ADMIN_PASS_ENC}@${DB_HOST}:${DB_PORT}/$1?sslmode=verify-full"
  export SRQL_TEST_DATABASE_SERVER_NAME="$TLS_SERVER_NAME"
  export SRQL_TEST_DATABASE_CA_CERT_FILE="$CA_FILE"
  export SERVICERADAR_TEST_DATABASE_QUEUE_TARGET_MS=10000
  export SERVICERADAR_TEST_DATABASE_QUEUE_INTERVAL_MS=10000
  export SERVICERADAR_TEST_SANDBOX_MODE=shared
  export MIX_ENV=test
}
```
</details>
- Outcome: ⚠️ 1 info across 2 runs (1h31m43s)

## Pipeline

Updates from [git push no-mistakes](https://github.com/kunchenguid/no-mistakes)

<details>
<summary>✅ **intent** - passed</summary>

✅ No issues found.
</details>

<details>
<summary>✅ **Rebase** - passed</summary>

✅ No issues found.
</details>

<details>
<summary>🔧 **Review** - 3 issues found → auto-fixed (2) ✅</summary>

- ⚠️ `elixir/serviceradar_core/lib/serviceradar/observability/stateful_alert_engine.ex:363` - The same Horde lookup lag the fix routes around is still consulted by ensure_started/1 (unchanged), so a drop remains reachable with concurrent callers (events/falco/trivy processors, log_promotion, liveness check, Oban workers all call the engine concurrently). Sequence: caller A start_child -&gt; P1 registers (register_name always replies :yes; Horde.RegistryImpl never checks for an existing key) but the keys ETS row is not written yet; caller B&#39;s ensure_started sees lookup == [] and start_child&#39;s a second engine P2 for the same key (Horde.DynamicSupervisor does not dedupe by child id); when the registry processes P2&#39;s diff, process_diff/2 does Process.exit(P1, {:name_conflict, ...}). If A&#39;s GenServer.call to P1 is in flight, it exits with {{:name_conflict, ...}, _}, which call_pid/2 maps to the generic {:error, reason} and does NOT retry (only :noproc retries), so A&#39;s batch is still dropped; meanwhile two engines with separate in-memory rule state briefly evaluate the same shard. This is pre-existing and the change strictly narrows the failure (A previously failed unconditionally in this window), so it does not block the fix. Closing it needs serialised shard starts or treating a name_conflict exit as retryable, which has double-evaluation implications for a partially processed batch - the remedy, not the defect, extends the change beyond its stated intent and needs authorization.
- ℹ️ `elixir/serviceradar_core/test/serviceradar/observability/stateful_alert_engine_test.exs:1609` - The inline comment says the second evaluate runs &#39;while the killed shard&#39;s registry state is still settling, so a name-based call would exit :noproc&#39;. The author&#39;s own verification (the test passes on the unfixed code) shows that is not what happens: after the :DOWN, Horde.Registry.lookup&#39;s process_alive? filter returns [], ensure_started starts a fresh shard, and a name-based call succeeds. The header comment above the test states the limitation correctly; this inline comment overstates what the test exercises and should be reworded to say it only pins restart-and-evaluate after an out-of-band termination.
- ℹ️ `elixir/serviceradar_core/lib/serviceradar/observability/stateful_alert_engine.ex:326` - The comment attributes the lookup lag to the key&#39;s member not being in the registry&#39;s &#39;locally materialized member set&#39;. The confirmed root cause in Horde 0.10 is different: handle_call({:register, ...}) replies right after DeltaCrdt.put, and the keys ETS row that Horde.Registry.lookup/2 reads is only inserted later by process_diff/2 when the asynchronous CRDT diff is handled. The member_in_cluster? check is a separate condition. Reword so the next reader is pointed at the real mechanism. Also, the rolling-deploy comment near line 307 still refers to the caller&#39;s `call/2` catch, which is now call_pid/2.

🔧 Fix applied.
1 warning still open:

- ⚠️ `elixir/serviceradar_core/test/serviceradar/observability/stateful_alert_engine_test.exs:1607` - The new test reads Horde&#39;s registry synchronously right after a shard was started, which is exactly the read the change documents as laggy. Line 1607 (`[{pid, _metadata}] = ProcessRegistry.lookup(:stateful_alert_engine)`) runs immediately after the first evaluate_events started shard 0 through start_child; line 1619 (`assert [{new_pid, _metadata}] = ProcessRegistry.lookup(...)`) runs immediately after the restarted shard was called BY PID. With the fix, evaluate_events no longer waits on (or depends on) the keys ETS row being written by process_diff/2, so under the registry contention seen on BazelCI integration_tests_serial_4 both lookups can return [] for a live shard: 1607 raises MatchError, 1619 fails the assertion. The fix makes the engine tolerant of the lag while the test that ships with it is not, so it can reintroduce the same CI flake it is meant to close. Wrap both lookups in the file&#39;s existing `eventually/2` helper (e.g. `eventually(fn -&gt; ProcessRegistry.lookup(:stateful_alert_engine) end, &amp;match?([{_, _}], &amp;1))`) before matching; no production change needed.

🔧 Fix applied.
✅ Re-checked - no issues remain.
</details>

<details>
<summary>⚠️ **Test** - 1 info</summary>

- ℹ️ `elixir/serviceradar_core/test/serviceradar/observability/stateful_alert_engine_test.exs:1546` - The intent says the lookup-lag race cannot be forced without churning the registry (which bloats the CRDT) or poking Horde&#39;s private ETS tables. A third way works and avoids both. Flood the registry with read-only `GenServer.call(registry, :members)` callers (2000 spawned processes) while a drained shard is started. `:members` is answered from registry state and writes nothing to the CRDT, but it queues ahead of the `{:crdt_update, _}` diff. With that in place the base code dropped 13/25 evaluate batches and 22/25 resolve sweeps, and the fixed code dropped 0/25 of each. The probe is saved as lag_probe_test.exs in the evidence directory. I did not add it to the committed suite because it contradicts the author&#39;s stated decision to pin only the deterministic half. Adopting it would give a regression test that fails before the fix and passes after it, and the header comment saying the race &#39;is not forced here&#39; would then need updating. It adds 2000 short-lived processes per iteration to a serial CI lane, so whether to adopt it is the author&#39;s call.
- Live validation: ✅ go - 4 of 5 scenarios driven live against the product

| Scenario | Result | Live | Evidence |
| --- | --- | --- | --- |
| An event batch sent to a freshly started shard while the Horde registry is busy is evaluated instead of returning {:error, :engine_not_running} | ✅ pass | live | probe-fixed.txt (25/25 :ok on the target commit) vs probe-base.txt (13/25 {:error, :engine_not_running} on the base commit). Same probe, same runtime, same database. |
| A stale-anomaly resolve sweep sent to a freshly started shard under registry contention returns {:ok, n} instead of {:error, :engine_not_running} | ✅ pass | live | probe-resolve-fixed.txt (25/25 {:ok, 0}) vs probe-resolve-base.txt (22/25 {:error, :engine_not_running}) |
| A shard terminated out-of-band is restarted and the next batch raises its alert (second alert row persisted, new pid registered) | ✅ pass | live | `mix test .../stateful_alert_engine_test.exs:1551 --include integration` against the real Horde supervisor and the CNPG scratch database: 1 test, 0 failures |
| Existing engine behaviour is unchanged: multi-shard fan-out, fire/resolve, cooldown dedupe, the resolve sweep and the liveness check still work through the new call path | ✅ pass | live | engine-test-file-fixed.txt: full stateful_alert_engine_test.exs, 15 tests, 0 failures on the target commit |
| Adversarial: a shard pid dies in the window between ensure_started/1 resolving it and the GenServer.call, and the single retry re-resolves and delivers the batch | ⏸️ untested | no | That window is microseconds wide, inside private functions, with no injection seam. Forcing it would need a production test hook or tracing-based process suspension, and neither is available without c… |

- <code>`MIX_ENV=test mix ecto.migrate` against a fresh `codex_nm4477_*` scratch database on the srql-fixtures CNPG cluster (480 migrations applied; needed `SERVICERADAR_TEST_DATABASE_OWNERSHIP_TIMEOUT_MS` raised because the default 120s sandbox ownership timeout kills the migration lock)</code>
- <code>`mix test test/serviceradar/observability/stateful_alert_engine_test.exs --include integration --trace` on target commit: 15 tests, 0 failures</code>
- <code>`mix test test/serviceradar/observability/stateful_alert_engine_test.exs:1551 --include integration` on target commit after restoring the fix: 1 test, 0 failures</code>
- <code>Temporary probe `lag_probe_test.exs` (25 iterations, 2000 read-only `:members` flooders) calling `StatefulAlertEngine.evaluate_events/1` on a freshly started shard. Target commit: 25/25 `:ok`. Base engine file (`git show 42f5f0b10a:...stateful_alert_engine.ex`): 13/25 `{:error, :engine_not_running}`</code>
- <code>Same probe with `NM_PROBE_MODE=resolve` calling `StatefulAlertEngine.resolve_stale_anomalies/3`. Target commit: 25/25 `{:ok, 0}`. Base engine file: 22/25 `{:error, :engine_not_running}`</code>
- <code>Cleanup checks: `git diff --quiet HEAD -- lib` confirmed the fix was restored, scratch database dropped and re-queried (0 matching databases), `git status --short --ignored` clean</code>

🔧 Fix applied.
1 info still open:

- ℹ️ `.agents/skills/srql-fixtures-db-tests/SKILL.md:118` - Not caused by this change and not blocking. The srql-fixtures-db-tests skill tells the reader to run `mix ecto.migrate` on a fresh scratch database, while AGENTS.md says to use `mix serviceradar.db.migrate`. Following the skill, a full replay against the remote fixture lost its migration lock after 326 migrations. A second `mix ecto.migrate` could not resume, because the ledger had been created in `public` before the `platform` schema existed, and Ecto then opened a new empty `platform.schema_migrations` and failed on already-existing tables. `mix serviceradar.db.migrate` on a fresh database worked first time. The skill text may be worth aligning with AGENTS.md in a separate change.
- Live validation: ✅ go - 5 of 7 scenarios driven live against the product

| Scenario | Result | Live | Evidence |
| --- | --- | --- | --- |
| An operator&#39;s event batch arrives at a shard that was just started while the registry is busy; the engine returns :ok and evaluates it instead of dropping it | ✅ pass | live | round2-fixed-full-file.txt (test &#39;a batch sent to a just-started shard is evaluated while the registry lookup lags&#39; passes, 12 iterations all :ok); round2-seed-stability.txt FIXED seeds 1-5 |
| The stale-anomaly resolve sweep reaches a just-started shard while the registry is busy; it returns {:ok, count} instead of an error | ✅ pass | live | round2-fixed-full-file.txt (test &#39;a resolve sweep sent to a just-started shard runs while the registry lookup lags&#39; passes, all {:ok, 0}); round2-seed-stability.txt FIXED seeds 1-5 |
| Adversarial: the same lookup-lag scenarios run against the unfixed name-based dispatch must reproduce the dropped batch, proving the regression tests can fail | ✅ pass | live | round2-unfixed-dispatch.txt (left: [:ok, {:error, :engine_not_running}] and [ok: 0, error: :engine_not_running]); round2-unfixed-detection-rate.txt (both tests fail 8/8) |
| A shard terminated out-of-band is restarted by the next batch, the batch produces its alert, and a new live pid is registered under the shard key | ✅ pass | live | round2-fixed-full-file.txt (test &#39;a shard terminated out-of-band is restarted and takes the next batch&#39;); also passes on base code in round2-unfixed-dispatch.txt, consistent with the author&#39;s declared… |
| Existing engine behaviour is unchanged: bucketed alerts, metric baselines, anomaly open/clear, shard fan-out, liveness check and dedup all still work | ✅ pass | live | round2-fixed-full-file.txt: 17 tests, 0 failures, MIX_EXIT=0 |
| The shard pid dies between ensure_started/1 and the GenServer.call; the single retry re-resolves and the batch is still evaluated | ⏸️ untested | no | The window between pid resolution and the call is inside a private function, and nothing available here can force a kill in that window. It would need a test hook in the dispatch path or a tracing-bas… |
| Under BazelCI&#39;s integration_tests_serial_4 lane, where the flake was first observed, the engine no longer drops freshly-started-shard batches | ⏸️ untested | no | The guarded Bazel integration lifecycle may only run in the in-cluster BuildBuddy workflow (typed SERVICERADAR_ENV=ci configuration and private fixture secrets); the project skill forbids running it f… |

- <code>`mix serviceradar.db.migrate` on a fresh scratch DB on srql-fixtures; verified MIX_EXIT=0, single ledger platform.schema_migrations with 480 rows vs 480 migrations on disk</code>
- <code>`mix test test/serviceradar/observability/stateful_alert_engine_test.exs --include integration --trace --seed 0` on HEAD 0f74b7c8b1 (pool_size 12): 17 tests, 0 failures</code>
- <code>`git checkout 42f5f0b10a -- lib/serviceradar/observability/stateful_alert_engine.ex` then `mix test &lt;file&gt;:1555 &lt;file&gt;:1650 &lt;file&gt;:1665 --include integration --trace --seed 0`: 3 tests, 2 failures (both lookup-lag tests), file restored afterwards</code>
- <code>Lookup-lag tests (`&lt;file&gt;:1650 &lt;file&gt;:1665`) on fixed code, seeds 1-5: 5/5 pass</code>
- `Lookup-lag tests on base-commit engine, seeds 11-18 with per-test failure counting: both tests failed in 8/8 runs`
- <code>Cleanup checks: `DROP DATABASE` of my scratch DB followed by a re-query showing 0 matching databases; `git status --short --untracked-files=all` empty; HEAD unchanged</code>
</details>

<details>
<summary>✅ **Document** - passed</summary>

✅ No issues found.
</details>

<details>
<summary>✅ **Lint** - passed</summary>

✅ No issues found.
</details>

<details>
<summary>✅ **Push** - passed</summary>

✅ No issues found.
</details>


mfreeman451 (commented):
lgtm

PR #4545, opened 2026-09-21
fix(core): stream managed-device enforcement instead of one keyset page

mfreeman451:
## Problem

`MtrAutomationDispatcher.enforce_managed_baseline_targets/1` called `Ash.read/2`
with no page options, relying on `Device.read`'s
`pagination keyset?: true, default_limit: 5000` to return every matching row in
a single page. That page silently comes back short.

Measured against a live deployment on the identical query and actor:

```
Ash.count!(query)                                  => 345
Ash.read!(query, page: [limit: 5000, count: true]) => 251 results
                                                      page count 345
                                                      more? false
```

The 5000 limit was never the binding constraint. Ash's own page count agrees
there are 345 rows, returns 251 of them, and reports `more? == false`, so no
caller can detect that 94 rows were dropped. Walking the same query in 50-row
pages returned all 345 distinct uids, and so does `Ash.stream!/2`.

Ruled out while narrowing this: the actor is `role: :system` so `system_bypass()`
applies and policies are not filtering; an explicit `sort(uid: :asc)` on the
unique primary key changes nothing; the missing rows are spread proportionally
across every device `type`, so it is not a predicate; and the SRQL side that
feeds the candidate list returns all 365 rows correctly.

## Why it was invisible

A device missing from the read simply fails the `device_map` lookup in
`filter_targets_by_devices/2`, and `device_allowed?(nil)` returns `false`. So
every lost row was treated as **unmanaged** and dropped from the baseline sweep.

The observed symptom was an MTR automation profile that traced 251 of 345
eligible devices on every run, with nothing logged and no indication in the
product that coverage was partial. Raising the profile's selector limit cannot
help, because the cap sits upstream of the limit.

## Fix

Use `Ash.stream!/2`, which pages internally, with a batch size well below the
resource's default page limit. The previous fail-open contract is preserved: a
read failure still logs a warning and returns the targets unfiltered rather than
aborting a sweep.

## Testing

Compiles clean with no new warnings, and `mix format --check-formatted` passes.
The existing `MtrAutomationDispatcherTest` is database-free and only covers
`classify_transition/2`, so it does not exercise this path.

**This change is verified against a live deployment, not by a new automated
test, and I want to be explicit about that.** Reproducing it requires a database
holding more than ~251 rows matching one filter, which is the integration-DB
path rather than a unit test. I am happy to add that integration test in this PR
if you want it gated that way; the honest caveat is that the truncation
threshold may be data- or version-dependent, so a smaller fixture could pass
both before and after the change and give false confidence.

Worth a maintainer's judgement separately: if a single large keyset page can
return short with `more? == false`, other bare `Ash.read/2` calls on
`default_limit: 5000` resources have the same exposure. `baseline_targets/1` in
this same module takes one page too, though it caps deliberately with
`Ash.Query.limit/2`, so it is bounded by intent rather than by accident.


mfreeman451 (commented):
lgtm

PR #4548, opened 2026-09-22
fix(inventory): close two remaining interface-mac-chassis merge gaps

mfreeman451:
## Summary

- **Reserved MAC guard**: `interface_mac_chassis_groups_from_rows/1` now rejects reserved MAC values (all-zeros, broadcast) in the pure function, mirroring the existing LAA guard. Pre-fix rows may still exist in `DeviceInterfaceMac`; this ensures no merge can be driven from them regardless of table contents.
- **Cross-partition guard**: Added `where: owner.partition == other.partition` to the `interface_mac_chassis_groups/0` query. The sibling source `column_mac_groups` already had this constraint; `interface_mac_chassis_groups` was missing it, allowing devices in different partitions to form a merge component.

Both gaps were identified during root-cause analysis of bad device merges (#4546). PR #4546 fixed the normalization chokepoint; this PR closes the two remaining surface areas.

## Test plan
- [ ] New `"never merges on a reserved MAC value"` test in `DuplicateSweep.interface_mac_chassis_groups_from_rows/1` describe block
- [ ] All existing `DuplicateSweepTest` tests continue to pass (26 tests, 0 failures)
- [ ] `make test` passes


mfreeman451 (commented):
lgtm

PR #4562, opened 2026-09-22
fix(inventory): apply MergePolicy check in DuplicateSweep classify step

mfreeman451:
## Summary

- `DuplicateSweep` called `MergeEngine.merge_devices/3` directly, bypassing the policy-aware path (`merge_conflicting_devices/4`) that already calls `MergePolicy.merge_allowed_for_matches?/1`. MAC-only, agent_id-only, and medium-confidence-only two-device components were therefore merged unattended regardless of policy.
- Fix: after the size split in `classify_duplicate_components/1`, further split by `MergePolicy.merge_allowed_for_matches?/1`. Policy-blocked components join the `blocked` list so they are counted in metrics, logged by `report_blocked_components/1`, and never passed to the merge stage.
- Matches are built from `component.evidence` as `{type, %{value: value}}` — the same shape `MergePolicy` already expects — so `:mac` evidence is caught by `mac_only_matches?`, hardware evidence types (`:mac_sibling`, `:mac_column`, `:interface_mac_chassis`) pass through.

## Test plan

- [ ] `DuplicateSweep.classify_duplicate_components/1` with `:mac`-only evidence pair → `blocked`
- [ ] Same with mixed `:mac` + `:agent_id` → `mergeable`
- [ ] Same with `:interface_mac_chassis` only → `mergeable` (hardware evidence, not blocked)
- [ ] `make test` Bazel unit shards green


mfreeman451 (commented):
lgtm

PR #4563, opened 2026-09-22
fix(inventory): DuplicateSweep policy gate + Device.read keyset sort

mfreeman451:
## Summary

- **DuplicateSweep policy gate**: `classify_duplicate_components/1` now calls `MergePolicy.merge_allowed_for_matches?/1` before marking a pair mergeable. Previously the policy split was only applied in the manual merge path; DuplicateSweep bypassed it entirely, allowing MAC-only matches to proceed as automatic merges.

- **Device.read keyset pagination truncation fix**: Adds `prepare build(sort: [uid: :asc])` to the `:read` action. Without an explicit ORDER BY, PostgreSQL returned fewer rows than `Ash.count!` for large-limit keyset pages (measured: 251 results with `more?: false` when count was 345). This is the root-cause fix; the streaming workaround in `mtr_automation_dispatcher.ex` (PR #4545) remains as defense-in-depth.

## Changes

`elixir/serviceradar_core/lib/serviceradar/inventory/identity/duplicate_sweep.ex`
- Alias `MergePolicy`
- Split `{policy_ok, policy_blocked}` in `classify_duplicate_components/1` using `merge_allowed_for_matches?/1`

`elixir/serviceradar_core/test/serviceradar/inventory/identity/duplicate_sweep_test.exs`
- 3 new behavioral tests: MAC-only evidence → blocked; MAC + agent_id → allowed; `:interface_mac_chassis` → allowed (not blocked by MAC-only rule)

`elixir/serviceradar_core/lib/serviceradar/inventory/device.ex`
- `prepare build(sort: [uid: :asc])` on `:read` action

`elixir/serviceradar_core/test/serviceradar/inventory/device_soft_delete_test.exs`
- Regression test: creates 8 tagged devices, reads back with `page: [limit: 100, count: true]`, asserts `length == count` and `more?: false`

## Test plan

- [ ] `bazel test //elixir/serviceradar_core:unit_tests_*` — DuplicateSweep + Device unit suites
- [ ] Integration shard covering `device_soft_delete_test.exs` confirms keyset pagination completeness
- [ ] Integration shard covering `duplicate_sweep_test.exs` confirms MAC-only block


mfreeman451:
Closing in favor of a clean, targeted fix for the keyset pagination issue only. This PR removes the reserved_mac_value? guard and partition equality check that protect against phantom device re-adoption — those regressions should not be included.

